PR #1, opened 2017-07-11
__init__.py to gnpy.py

giladg-FB:
Changing __init__.py to gnpy.py; python packaging requires specific naming conventions and structure.


ggrammel:
Hi Gilad,

Thank you. I copied it to my local pc and ran python3 gnpy. Worked OK in sense that the program was executed and an output was produced. Unfortunately the computed GNLI re not visible there but I guess that is not on your plate to fix. Normally I would file an issue in Github, but given the 2week code-freeze period we are still in, it may not be adequate. I"ll send a separate email to Vittorio so he is aware. Probably they already have a fix and are just waiting for us to un-block the freeze.

Gert

[cid:image001.png@01D2FB0A.4E1B4360]

PR #29, opened 2018-02-19
code wrap up and edfa model augmentation

jeanluc-auge:
-new edfa_config.json format withgain_min, gain_max, p_max and a 2nd edfa model in case the nf polynomial fit is not known: nf_model field with true/false enabled parameter. Takes ofr input the nf_min @gain_min and nf_max@gain_max
-build_oa_json.py generate the edfa_config.json file from OA.json file and precalcualtes the nf_model parameters
-implement the 2nd edfa model in elemtes.py and check for edfa power violation and min_gain.


jeanluc-auge:
closing it to remove 2 unwanted files (amplifier2.py and amplifier_pytest.py

PR #33, opened 2018-02-26
Add NLI Evaluation in Fiber class

mcantono:
Add methods to compute NLI via the Incoherent Gaussian Noise Model in Fiber element, integrating @jeanluc-auge PR
Add Documentation for the Gaussian Noise model


mcantono:
With respect to unit of measures, we rolled back to SI-based units. We also made few changes in the Fiber methods, putting them as properties rather than explicit methods.

PR #44, opened 2018-03-26
Add source sink arguments, restore convert.py

jeanluc-auge:
demonstrate mesh routing capability with source & sink options
python transmission_main_example.py coronet_conus_example.json Toledo Syracuse
=> path feasibility from Toledo to Syracuse
python transmission_main_example.py coronet_conus_example.json -l
=> list all possible source and destination nodes
readd CORONET xls and convert.py parser from xls to json

Signed-off-by: Jean-Luc Auge <jeanluc.auge@orange.com>


examples/transmission_main_example.py
```diff
@@ -47,40 +47,55 @@ def main(args):
...
+        axis('off')
+        show()
 
 parser = ArgumentParser()
 parser.add_argument('filename', nargs='?', type=Path,
   default= Path(__file__).parent / 'edfa/edfa_example_network.json')
+parser.add_argument('source', type=str, nargs='?', default="", help='source node')
+parser.add_argument('sink', type=str, nargs='?', default="", help='sink node')
 parser.add_argument('-v', '--verbose', action='count')
+parser.add_argument('-l', '--list', action='count', default=0, help='list all network nodes')
```
dutc: This should be `action='store_true'` rather than `action='count'`.

Can you call this parameter "list-nodes" to make it clear what it is listing?

You can check if the parameter was passed by simply
`
if args.list_nodes:
  ...
`

examples/transmission_main_example.py
```diff
@@ -47,40 +47,55 @@ def main(args):
...
-    fig = figure()
-    pos = {n: (n.lng, n.lat) for n in nodelist}
-    kwargs = {'figure': fig, 'pos': pos}
-    plot = draw_networkx_nodes(network, nodelist=nodelist, node_color=node_color, **kwargs)
-    draw_networkx_edges(network, edgelist=edgelist, edge_color=edge_color, **kwargs)
-    draw_networkx_labels(network, labels=labels, font_size=14, **kwargs)
-    title(f'Propagating from {source.loc.city} to {sink.loc.city}')
-    axis('off')
-    show()
+    if args.list>=1:
```
dutc: I don't think listing all the network nodes should be exclusive with running a simulation.

If that's the case, there should be a greater separation between these two modalties.

examples/transmission_main_example.py
```diff
@@ -47,40 +47,55 @@ def main(args):
...
+        for el in trx:
+            print(el.uid[4:])
+    else:
+        source, sink = trx[0], trx[1]
+        for el in trx:
```
dutc: You dont account for the possibility that the user did not supply a source and sink agument, other than by specifying an invalid default value `""`.

Either:
- make these parameters completely optional and dynamically determine a reasonable default value (which is not particularly meaningful,)
- make these values mandatory.

dutc: Additionally, you also don't account for the possibility that the source and sink are invalid.

For now, it's okay that you're assuming that the node uids encode the network element type in the first four characters, but that may change later (in which case we'll have to fix this code.)

I would break this into two separate loops which `break` once they have identified the correct element, plus a check at the end.

You could write:
```
source = None
for el in trx:
    if args.source == el.uid[4:]:
        source = el
        break
if source is None:
    raise ValueError(f"invalid source node specified: {args.source!r}")
```

You could simplify to:
```
source = next((el for el in trx if args.source == el.uid[4:]), None)
if source is None:
    raise ValueError(f"invalid source node specified: {args.source!r}")
```

Or even:
```
try:
    source = next(el for el in trx if args.source == el.uid[4:])
except StopIteration as e:
    raise ValueError(f"invalid source node specified: {args.source!r}") from e
```

dutc: Also, try to keep the most common code path at the lowest level of indentation.

As you can see, by intenting the block below, the diff for this PR is much larger than it should be.

PR #48, opened 2018-04-06
Advanced xls support, eqpt library and mesh routing

jeanluc-auge:
please don't squash the commits, thanks


jeanluc-auge:
-source and destination nodes can be entered in parameters. Default = 2 first nodes in the network topology
-edfa and fibers characteristics are defined in eqpt_config.json (last part of the file on spans, roadm and Si is not yet implemented). => eqpt are reffered by their name, for example 'std_medium_gain' or 'SSMF'
-accepts an xls or a json file as input. If an xls file is input in parameter, the corresponding json will be generated (same name as the xls file) and the simulation is run from the json file.
-increased parsing capabilities of the xls file: see meshTopologyExampleV2.xls
-full backward compatbility with CoronetConus xls network. It can still be parsed, even fields are missing : default values are provided. They are visible in the json output

dutc:
I re-added the missing coronet files. They may have gotten lost doing some directory cleanup.

I re-added them to their own subfolder in examples/

In general, examples/ should clearly keep the different example code separated.

Can you re-pull and adjust your commit to reflect this?

Also, can you address the issues in the review?

examples/eqpt_config.json
```diff
@@ -0,0 +1,43 @@
+{     "Edfa":[{
+            "type_variety": "std_medium_gain",
```
dutc: Can we pick a better name than "type_variety"?

examples/eqpt_config.json
```diff
@@ -0,0 +1,43 @@
...
+            "gain_flatmax": 25,
+            "gain_min": 15,
+            "p_max": 21,
+            "advanced_config_from_json": "std_medium_gain_advanced_config.json"
```
dutc: Is there a distinction between "advanced config" and another type of config?

Ideally, we don't want one file pointing to other files. Ideally, we have a much flatter tree with the network config points to all of these files directly.

examples/eqpt_config.json
dutc: Can we also pick a better name than "egpt_config.json"?

examples/eqpt_config.json
```diff
@@ -0,0 +1,43 @@
...
+            {
+            "type_variety": "std_low_gain",
+            "gain_flatmax": 19,
+            "gain_min": 12,
+            "p_max": 22,
+            "nf_min": 6.5,
+            "nf_max": 10
+            }
+      ],
+      "Fiber":[{
```
dutc: I presume these are arrays of maps, so that you can specify subtypes of "Fiber."

Could this be a simple map of maps?

```
"Fiber": { "SSMF": { ... }, ... }
```

examples/transmission_main_example.py
```diff
@@ -77,10 +61,63 @@ def main(args):
     axis('off')
     show()
 
+
+def main(args):
+    input_filename = str(args.filename)
+    split_filename = input_filename.split(".")
+    json_filename = split_filename[0]+'.json'
```
dutc: `args.filename` is a `Path` object. Prefer use of native operations rather than coercion to string and use of string manipulation.

e.g., 
```
p = Path('a/b/c/d.txt')
p.with_suffix('.json')
```

examples/transmission_main_example.py
```diff
@@ -77,10 +61,63 @@ def main(args):
...
+    try:
+        assert split_filename[1] in ('json','xls','csv','xlsm')
```
dutc: When you run `python -O` (with optimisations,) assertions are elided from the AST.

What this means is that `assert` is a form of documentation, not a way to check business logic.

Instead of:
```
try:
  assert pred
except AssertionError as e:
  raise pred
```

Do:
```
if not pred:
   raise Exception(...)
```

Use `assert` to communication information to the person reading the code, not to perform critical validations.

dutc: Furthermore, this logic can be wrapped into the `ArgumentParser`.

Set the `type` to `json_path`:
```
from argparse import ArgumentParser
from pathlib import Path

parser = ArgumentParser()
def json_path(p):
    p = Path(p)
    if p.suffix != '.json':
        parser.error('must be .json file')
    return p
parser.add_argument('filename', type=json_path)

args = parser.parse_args(['abc.txt'])
print(args)

args = parser.parse_args(['abc.json'])
print(args)
```

examples/transmission_main_example.py
```diff
@@ -77,10 +61,63 @@ def main(args):
...
+    except AssertionError as e:
+        print(f'invalid file extension .{split_filename[1]}')
+        raise e
+    if split_filename[1] != 'json':
+        print(f'parse excel input to {json_filename}')
+        convert_file(input_filename)
+
+    json_data = load_json(json_filename)
+    read_eqpt_library(EQPT_LIBRARY_FILENAME)
```
dutc: We probably don't need `eqpt_library` to be a global variable here.

Prefer to keep state management very direct:
`equipment = read_equipment(filename)`

Rather than hardcoding the `EQPT_LIBRARY_FILENAME` path, this can be an argument provided to the argument parser with a default.

examples/transmission_main_example.py
```diff
@@ -77,10 +61,63 @@ def main(args):
...
+    network = network_from_json(json_data)
+    build_network(network)
+
+    spacing = 0.05 #THz
+    si = SpectralInformation() # !! SI units W, Hz
+    si = si.update(carriers=tuple(Channel(f, (191.3+spacing*f)*1e12, 
+            32e9, 0.15, Power(1e-3, 0, 0)) for f in range(1,97)))
+
+    trx = [n for n in network.nodes() if isinstance(n, Transceiver)]
+    if args.list>=1:
```
dutc: (Copied from previous PR review):

I don't think listing all the network nodes should be exclusive with running a simulation.

If that's the case, there should be a greater separation between these two modalties.

Also, if this is the case, return early rather than indenting everything below.

examples/transmission_main_example.py
```diff
@@ -77,10 +61,63 @@ def main(args):
...
+        print(*[el.uid for el in trx], sep='\n')
```
dutc: You dont account for the possibility that the user did not supply a source and sink agument, other than by specifying an invalid default value `""`.

Either:
- make these parameters completely optional and dynamically determine a reasonable default value (which is not particularly meaningful,)
- make these values mandatory.

dutc: Additionally, you also don't account for the possibility that the source and sink are invalid.

For now, it's okay that you're assuming that the node uids encode the network element type in the first four characters, but that may change later (in which case we'll have to fix this code.)

I would break this into two separate loops which `break` once they have identified the correct element, plus a check at the end.

You could write:
```
source = None
for el in trx:
    if args.source == el.uid[4:]:
        source = el
        break
if source is None:
    raise ValueError(f"invalid source node specified: {args.source!r}")
```

You could simplify to:
```
source = next((el for el in trx if args.source == el.uid[4:]), None)
if source is None:
    raise ValueError(f"invalid source node specified: {args.source!r}")
```

Or even:
```
try:
    source = next(el for el in trx if args.source == el.uid[4:])
except StopIteration as e:
    raise ValueError(f"invalid source node specified: {args.source!r}") from e
```

examples/transmission_main_example.py
```diff
@@ -77,10 +61,63 @@ def main(args):
...
+            print(f'invalid destination node specified: {args.sink!r}, replaced with {sink}')
+
+        path = dijkstra_path(network, source, sink)
+        print(f'There are {len(path)} network elements between {source} and {sink}')
+
+        for el in path:
+            si = el(si)
+            print(el)
+
+        #plot_network_graph(network, path, source, sink)
```
dutc: We can make this a flag enabled or disabled in the parser arguments below.

examples/transmission_main_example.py
```diff
@@ -77,10 +61,63 @@ def main(args):
...
+
 parser = ArgumentParser()
 parser.add_argument('filename', nargs='?', type=Path,
-  default= Path(__file__).parent / 'edfa/edfa_example_network.json')
+  default= Path(__file__).parent / 'edfa_example_network.json')
+parser.add_argument('source', type=str, nargs='?', default="", help='source node')
+parser.add_argument('sink', type=str, nargs='?', default="", help='sink node')
 parser.add_argument('-v', '--verbose', action='count')
+parser.add_argument('-l', '--list', action='count', default=0, help='list all network nodes')
```
dutc: This should be `action='store_true'` rather than `action='count'`.

Can you call this parameter "list-nodes" to make it clear what it is listing?

You can check if the parameter was passed by simply
`
if args.list_nodes:
  ...
`

gnpy/core/elements.py
```diff
@@ -246,6 +247,7 @@ def __str__(self):
             return f'{type(self).__name__} {self.uid}'
 
         return '\n'.join([f'{type(self).__name__} {self.uid}',
+                          f'  type_variety:      {self.config.type_variety}',
```
dutc: Is it `self.type_variety` or `self.config.type_variety`?

gnpy/core/equipment.py
```diff
@@ -0,0 +1,119 @@
...
+
+'''
+nf model parameters calculation
+calculate nf1, nf2 and Delta_P of a 2 coils edfa with internal VOA
+from nf_min and nf_max inputs 
+'''
+import numpy as np
+from gnpy.core.utils import lin2db, db2lin, load_json
+
+GAIN_MIN_FIELD = "gain_min"
```
dutc: Unless someone is going to change this at runtime, I don't think we gain anything from using a constant here.

It is not uncommon for Python code to use strings as constants.

In this particular case, you're basically just trading a `NameError` for a `KeyError`.

If you like the use of enumerations, there is https://docs.python.org/3/library/enum.html

However, in this case, field keys are generally just strings.

gnpy/core/equipment.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def nf_model(amp_dict):
+    gain_min = amp_dict[GAIN_MIN_FIELD]
+    gain_max = amp_dict[GAIN_MAX_FIELD]
+    try:
+        nf_min = amp_dict.get(NF_MIN_FIELD,-100)
+        nf_max = amp_dict.get(NF_MAX_FIELD,-100)
+        if nf_min<-10 or nf_max<-10:
+            raise ValueError
```
dutc: The only thing in this block that could raise the `ValueError` is this explicit `raise`, but you handle it immediately below.

I think this can be simplified to:
```
if nf_min < -10 or nf_max < -10:
   print( ... )
```

gnpy/core/equipment.py
```diff
@@ -0,0 +1,119 @@
...
+    g1a_max = gain_max - delta_p
+    #nf1 and nf2 are the nf of the 1st and 2nd stage coils
+    #calculate nf1 and nf2 values that solve nf_[min/max] = nf1 + nf2 / g1a[min/max]
+    nf2 = lin2db((db2lin(nf_min) - db2lin(nf_max)) / (1/db2lin(g1a_max)-1/db2lin(g1a_min)))
+    nf1 = lin2db(db2lin(nf_min)- db2lin(nf2)/db2lin(g1a_max)) #expression (1)
+
+    """ now checking and recalculating the results:
+    recalculate delta_p to check it is within [1-6] boundaries
+    This is to check that the nf_min and nf_max values from the json file
+    make sense. If not a warning is printed """
```
dutc: This should probably be a comment rather than a docstring.

gnpy/core/equipment.py
```diff
@@ -0,0 +1,119 @@
...
+            print('1st coil vs 2nd coil calculated DeltaP {} is not valid: revise inputs'
+                        .format(delta_p))
+    #check the calculated values for nf1 & nf2:
+    nf_min_calc = lin2db(db2lin(nf1) + db2lin(nf2)/db2lin(g1a_max))
+    nf_max_calc = lin2db(db2lin(nf1) + db2lin(nf2)/db2lin(g1a_min))
+    if (abs(nf_min_calc-nf_min) > 0.01) or (abs(nf_max_calc-nf_max) > 0.01):
+        print('nf model calculation failed with nf_min {} and nf_max {} calculated'
+                .format(nf_min_calc, nf_max_calc))
+
+    return (nf1, nf2, delta_p)
```
dutc: You can simply:
`return nf1, nf2, delta_p`

gnpy/core/equipment.py
```diff
@@ -0,0 +1,119 @@
...
+    #calculate nf1 and nf2 values that solve nf_[min/max] = nf1 + nf2 / g1a[min/max]
+    nf2 = lin2db((db2lin(nf_min) - db2lin(nf_max)) / (1/db2lin(g1a_max)-1/db2lin(g1a_min)))
+    nf1 = lin2db(db2lin(nf_min)- db2lin(nf2)/db2lin(g1a_max)) #expression (1)
+
+    """ now checking and recalculating the results:
+    recalculate delta_p to check it is within [1-6] boundaries
+    This is to check that the nf_min and nf_max values from the json file
+    make sense. If not a warning is printed """
+    if nf1 < 4:
+        print('1st coil nf calculated value {} is too low: revise inputs'.format(nf1))
```
dutc: Should we raise a `ValueError` here?

gnpy/core/equipment.py
```diff
@@ -0,0 +1,119 @@
...
+    if nf2 < nf1 + 0.3 or nf2 > nf1 + 2: 
+        """nf2 should be with [nf1+0.5 - nf1 +2] boundaries
+        there shouldn't be very high nf differences between 2 coils
+        => recalculate delta_p 
+        """
```
dutc: This should probably be a comment rather than a docstring.

gnpy/core/equipment.py
```diff
@@ -0,0 +1,119 @@
...
+        """            
+        nf2 = max(nf2, nf1+0.3)
+        nf2 = min(nf2, nf1+2)
+        g1a_max = lin2db(db2lin(nf2) / (db2lin(nf_min) - db2lin(nf1))) #use expression (1)
+        delta_p = gain_max - g1a_max
+        g1a_min = gain_min - (gain_max-gain_min) - delta_p
+        if delta_p < 1 or delta_p > 6:
+            #delta_p should be > 1dB and < 6dB => consider user warning if not
+            print('1st coil vs 2nd coil calculated DeltaP {} is not valid: revise inputs'
+                        .format(delta_p))
```
dutc: Should we also raise a `ValueError` here?

Can we structure these two checks to move them to the top of the function rather than checking things after already performing computations?

gnpy/core/equipment.py
```diff
@@ -0,0 +1,119 @@
...
+    #check the calculated values for nf1 & nf2:
+    nf_min_calc = lin2db(db2lin(nf1) + db2lin(nf2)/db2lin(g1a_max))
+    nf_max_calc = lin2db(db2lin(nf1) + db2lin(nf2)/db2lin(g1a_min))
+    if (abs(nf_min_calc-nf_min) > 0.01) or (abs(nf_max_calc-nf_max) > 0.01):
+        print('nf model calculation failed with nf_min {} and nf_max {} calculated'
+                .format(nf_min_calc, nf_max_calc))
```
dutc: Same as above. We should avoid `print()` unless it's explicitly for desired user output. 

(Note that this an error, but it's sending to `stdout`…)

If it's an error, `raise Exception`. 

If it's not an error, you can use `warnings`: https://docs.python.org/3/library/warnings.html

If it's runtime logging, use `logging`: https://docs.python.org/3/library/logging.html

gnpy/core/equipment.py
```diff
@@ -0,0 +1,119 @@
...
+
+def read_eqpt_library(filename):
+    """build global dictionnary eqpt_library that stores all eqpt characteristics:
+    edfa type type_variety, fiber type_variety
+    from the eqpt_config.json (filename parameter)
+    also read advanced_config_from_json file parameters for edfa if they are available:
+    typically nf_ripple, dfg gain ripple, dgt and nf polynomial nf_fit_coeff
+    if advanced_config_from_json file parameter is not present: use nf_model:
+    requires nf_min and nf_max values boundaries of the edfa gain range
+    """
```
dutc: Could you revise this docstring to conform either to Sphinx or to PEP-257?

https://pythonhosted.org/an_example_pypi_project/sphinx.html#function-definitions

https://www.python.org/dev/peps/pep-0257/

gnpy/core/equipment.py
```diff
@@ -0,0 +1,119 @@
...
+    global eqpt_library
+    eqpt_library = load_json(filename)
```
dutc: As in previous comment, this function can return a value rather than mutating global state.

gnpy/core/equipment.py
```diff
@@ -0,0 +1,119 @@
...
+            #use a default ripple model (only default dgt is defined)
+            config_json_filename = DEFAULT_CONFIG_JSON_FILENAME
+            (nf1, nf2, delta_p) = nf_model(el)
+            #remove nf_min and nf_max field and replace by nf1, nf2 & delta_p
+            nf_min = el.pop('nf_min','')
+            nf_max = el.pop('nf_max','')
+            dict_nf_model['nf_model'] = dict(zip(["enabled","nf1","nf2","delta_p"],[True,nf1,nf2,delta_p]))
+
+        json_data = load_json(config_json_filename)
+        eqpt_library['Edfa'][i] = {**el, **json_data, **dict_nf_model}
```
dutc: It's probably better to return a new structure computed from the old structure rather than trying to change the old structure in-line.

Instead of:
```
obj = ...
for i, el in enumerate(obj):
  obj[i] = ...
```

You can do:
```
raw = ...
clean = ...
for el in raw:
   clean.append( ... ) # or clean[...] = ...
return clean
```

gnpy/core/equipment.py
```diff
@@ -0,0 +1,119 @@
...
+            config_json_filename = el.pop(ADVANCED_CONFIG_JSON_FILENAME)
+            dict_nf_model['nf_model'] = {'enabled': False}
+        else:
+            #use a default ripple model (only default dgt is defined)
+            config_json_filename = DEFAULT_CONFIG_JSON_FILENAME
+            (nf1, nf2, delta_p) = nf_model(el)
+            #remove nf_min and nf_max field and replace by nf1, nf2 & delta_p
+            nf_min = el.pop('nf_min','')
+            nf_max = el.pop('nf_max','')
+            dict_nf_model['nf_model'] = dict(zip(["enabled","nf1","nf2","delta_p"],[True,nf1,nf2,delta_p]))
```
dutc: It's probably easier to just use a literal here: `{'enabled': True, 'nf1': nf1, 'nf2': nf2, ...}`

gnpy/core/equipment.py
```diff
@@ -0,0 +1,119 @@
...
+        #go through the eqpt library to find the demanded type_variety
+        eqpt = next(eqpt for eqpt_type in eqpt_library
+                    for eqpt in eqpt_library[eqpt_type] 
+                    if eqpt.get('type_variety','') == eqpt_name)
+        eqpt_config = dict(eqpt)
+        del eqpt_config['type_variety']
+    except StopIteration as e:
+        print(f'cannot find eqpt {eqpt_name} in eqpt library')
+        #TODO log it
+    return eqpt_config
```
dutc: This whole lookup should be unnecessary.

The equipment library should just be a dict from which you can query the equipment you want by type/subtype.

gnpy/core/network.py
```diff
@@ -13,6 +13,7 @@
 from gnpy.core import elements
 from gnpy.core.elements import Fiber, Edfa, Transceiver, Roadm
 from gnpy.core.units import UNITS
+from gnpy.core.equipment import *
```
dutc: We should avoid `import *` in all contexts except possibly in `__init__.py`.

The use of `import *` is controversial in `__init__.py`, and we may also try to avoid it there.

If importing the individual items, as in `from gnpy.core.elements import ...` (above) is too tedious, you can `import gnpy.core.equipment as equipment` and then `equipment.xyz`

gnpy/core/network.py
```diff
@@ -92,6 +90,9 @@ def split_fiber(network, fiber):
     network = add_egress_amplifier(network, prev_node)
     return network
 
+def select_edfa(ingress_span_loss):
+    return "std_medium_gain"
```
dutc: `select_edfa` does not seem to be dependent on `ingress_span_loss`.

What are the other possibilities for return values?

Does this need to be a function right now?

examples/transmission_main_example.py
```diff
@@ -77,10 +61,63 @@ def main(args):
...
+
+    trx = [n for n in network.nodes() if isinstance(n, Transceiver)]
+    if args.list>=1:
+        print(*[el.uid for el in trx], sep='\n')
+    else:
+        try:
+            source = next(el for el in trx if el.uid == args.source)
+        except StopIteration as e:
+            source = trx[0]
+            print(f'invalid souce node specified: {args.source!r}, replaced with {source}')
```
dutc: Replace `print` to `stdout` with either `logging` or `warning`
https://docs.python.org/3/library/logging.html
https://docs.python.org/3/library/warnings.html

gnpy/core/equipment.py
jeanluc-auge: Strongly disagree.
If the topology parser calls for an eqpt that doesn't exist it should be logged out
=> it will be a very common mistake (spelling, libraires differences...)

examples/eqpt_config.json
jeanluc-auge: Just keeping the same convention than for the edfa_eqpt config and topology json

examples/eqpt_config.json
jeanluc-auge: presented and discussed in previous meetings.

gnpy/core/network.py
jeanluc-auge: Easy,
work in progress here...

examples/transmission_main_example.py
jeanluc-auge: yes I know: hence the #TODO logging in several parts of my code, but one thing at a time...
can you do it?

examples/transmission_main_example.py
jeanluc-auge: no, we don't want to read a possibly very large json file everytime we look up for an eqpt.

examples/transmission_main_example.py
jeanluc-auge: Sorry Jame, I don't find this comment helpful, 
especially as it is the 2nd time I'm pulling this part of the code.
You certainly don't want to list network nodes AND run the simulation. You are just looking for the nodes that will be the source and destination or your simulation.

examples/transmission_main_example.py
jeanluc-auge: Also not finding this useful,
the code you put is from previous PR
the new code accounts for missing or wrong value by providing a default node[0] and node(1] source and sink. (Like the previous PR did by the way...). And logging (printing) to the user that default values were used. These default values are fine for an example to the user and have the advantage to always exist because they are the 2 1st nodes of the network topology.

gnpy/core/elements.py
jeanluc-auge: the code works...

jeanluc-auge: James, 
the code works...

gnpy/core/equipment.py
jeanluc-auge: ?
These are just Cte to record the json field names, rather than hard coding in the middle of the code. This way if we change the attributs in the json eqpt file, it's easier to know what to modify in the code.

gnpy/core/equipment.py
jeanluc-auge: I want to stop the code execution.

gnpy/core/equipment.py
jeanluc-auge: no, not critical, unlike the ValueError above.

gnpy/core/equipment.py
dutc: It's fine to use hardcoded strings in this fashion.

Consider that the use a variable here adds a non-semantic, non-abstracted indirection, which is evident from the variable name being identical to its value.

gnpy/core/equipment.py
dutc: Okay, maybe we should remove the `except ValueError` so that the exception will terminate execution of this function.

gnpy/core/equipment.py
dutc: Okay, we should use `warnings` or `logging` for this, depending on what the precise nature of the problem.

examples/transmission_main_example.py
dutc: We should be able to load the equipment information just once while still avoiding the use of global state.

examples/transmission_main_example.py
dutc: Okay, we can keep these flags exclusive with each other.

We should push this modality out of this particular function into the `if __name__ == "__main__"` block.

gnpy/core/equipment.py
jeanluc-auge: By the way, this part of the code was already existing,
it's just been moved. My point is shouldn't we focus on what is new in this PR?
And yes it has to be there because they are the result of the 1st computation part. This is a sanity check of the 1st computation.

gnpy/core/equipment.py
jeanluc-auge: Likewise: existing code, not new to this PR.
But point taken for future code cleaning.

gnpy/core/equipment.py
jeanluc-auge: Likewise: existing code, not new to this PR.

examples/eqpt_config.json
jeanluc-auge: not an easy one, I'm trying to make it as clear as possible, suggestions are welcome.
For now there is a "advanced_config_from_json": parameter to read the detailed nf and ripple  model of Ciena from a user input json file that is specific to each amplifier variety_type.
If this advanced_config_from_json is not present in the edfa variety type definition, then a more basic model is used that use a standard (=the same for all amplifiers) input json file.It is hidden in the code so the user doesn't have to worry about it. If wanted the user to be able to input quickly a new edfa variety type without any measurment, nor json file to provide.

examples/transmission_main_example.py
jeanluc-auge: ok, please help here:
I'd like to finish the eqpt part of the code. Maybe you can modify this afterwards?

examples/transmission_main_example.py
jeanluc-auge: Please help, 
the global variable is the best I could find for this purpose.

examples/transmission_main_example.py
jeanluc-auge: ok,
for future improvement

examples/transmission_main_example.py
jeanluc-auge: Sorry, I had to put it in comments because it was not working when lattitude and longitude values = 0 were provided for all network elements (which is now the default values if they are not user provided).
=> Quick fix would be to make a check for longitude and lattitude values before launching the plot network graph.

examples/transmission_main_example.py
jeanluc-auge: sorry, not a parser pro, even though I read the doc.
Please implement and I can learn from it ;)

gnpy/core/equipment.py
jeanluc-auge: ok, will put the print in the if and remove the try/except bloc.

gnpy/core/equipment.py
jeanluc-auge: done

gnpy/core/equipment.py
jeanluc-auge: done

gnpy/core/equipment.py
jeanluc-auge: I have added WARNING in the print.
For the logging, let's implement it in the future.

gnpy/core/equipment.py
jeanluc-auge: done

gnpy/core/equipment.py
jeanluc-auge: raise valueError, 
done

gnpy/core/network.py
jeanluc-auge: done, 
it was a remain of my coding investigations

gnpy/core/network.py
luismartingarcia: @jeanluc-auge, this is a great addition to gnpy, thanks a lot for the hard work here! great stuff.

PR #70, opened 2018-05-02
Useability improvements 2

EstherLerouzic:
Dear James,
here is my pull request.
It mainly concerns improvment for excel file users.
It includes:
- one additional program to help formating the excel input file (create_eqpt_sheet.py)
- one additional help file (Excel_userguide.rst) concerning the format of the excel file and the use of this program
- some additional comments in the README file
regards

Esther


Excel_userguide.rst
```diff
@@ -0,0 +1,180 @@
+
```
dutc: This is good. Needs a few typos fixed which we can do in the next commits.

Would like you to think about how we can place this information in a few separate places (and easily keep them in sync):
- the Excel file itself (there should be a sample.xls that is a "self-documenting" example)
- the README & maybe the Github wiki

EstherLerouzic: Thanks for your comments !
Concerning your suggestions , the excel file meshTopologyExampleV2.xls is a good sample file. I'll try to create a sample file based on it ; I already put a link to the excel userguide in the README, but maybe this is too discreet?  adding this information on the wiki could help for sure. I'll  add this on the (long) TODO list :)

PR #80, opened 2018-06-12
Adding service list

EstherLerouzic:
I open this pull request to add some functions related to multiple request computation.
Merging with  the current develop is finished. 

this pull request contains:
- convert_service_sheet.py : xls parser, that can be called to create a json request file in accordance with:    Yang model for requesting Path Computation     draft-ietf-teas-yang-path-computation-01.txt. 
- path_requests_run.py : read json request file in accordance with the same yang model and returns path results in terms of path and feasibility
- an update of eqpt_config.json to include Transceivers types and modes
- an update of the service sheet from meshTopologyExampleV2.xls with all new required columns
- an update of the excel_userguide
- template for service and path_results json

Additionnally, the eqpt library can be entered as a parameter, and a specific spectral information is used for each request based on the transponder type and mode.

@ James please let me know if this description is OK .


dutc:
Superceded by PR #80

PR #84, opened 2018-06-29
Fix defaults

dutc:
Sample for coders onboarding call.


dutc:
This PR was for the coders onboarding call only.

PR #87, opened 2018-07-04
Connector att implem

EstherLerouzic:
These commits implement fiber spans connector losses.
- connector losses can be part of the json file or not. if not present default value is 0.0 (changed)
- automatic gain setup of edfa compensates the whole span loss inc connector losses
- test on json generation of files have been updated accordingly
- additional test to check OSNR consistency

Added on 09/07:
- "merging" of path_requests_run and transmission_main_example common functions (startingat commit #b9518ca...
   the aim is to maintain a single propagate functions. 
   Common classes and functions have been added to "core" directory : to be discussed !!
   with this structure: transmission_main_example includes the transponder type and parameter for the computation, and path_request_run uses the same propagate function.
 
-> missing (not finished): 
- additional test to check backward compatibility with files not having the connector loss attribute.


dutc:
I fixed the merge conflicts in this and merged with @jeanluc-auge's work in #89 

I will close this PR since #89 will include all of these commits and all of this work when it is accepted.

PR #91, opened 2018-07-09
network and equipment JSON descriptions added

berahtlv:
New library functions are added, some of them are optional. Human readable description of correct network and equipment JSON file format and its combination is needed.

Description is made against 'develop' branch. Maybe something is forgotten.

File extension is discussable.


dutc:
Thank you for this. This will be very useful. 

Could we provide this document in plaintext (Markdown)?

berahtlv:
Yes, but it will take some time.
I see that library is using .rst, however there is some extension preference?

dutc:
Yes, you're right. We should provide this as reStructured Text to file to match the other documentation.

I can help you with the conversion if you'd like. I have some tools for doing this quickly. Let me know, and I can send you a PR.

berahtlv:
Yes, it will be easier to convert than learn syntax:)

berahtlv:
Ok, I've made .rst file. 
This PR should be rejected and new one created or it is done differently?

PR #96, opened 2018-07-30
fixed gain amplifier model and default values implementation from eqpt_config json

jeanluc-auge:
_UPDATED DESCRIPTION AND COMMIT HISTORY:_
_code is now ready to be pulled and merged_

## __NEW__ amplifier model as "fixed_gain" 
* nf = Cte = nf0 and flat spectrum for gain in[gain_min, gain_flatmax] range
*nf=nf0 and tilted extended gain range for gain > gain_max
* automatic input padding if operational gain < min_gain
* smooth handling of user input mistakes in eqpt_config json

## __NEW__ small spans automatic padding
* The padding value is defined in eqpt_config: the feature is disabled if the value is set to 0
* Fiber spans with loss < padding are complemented with a fixed attenuator at the fiber input.
* If spans are Fused (spliced) together, the total loss is considered for padding and only the attenuator placed at the input of the 1st fiber
* A new att_in attribute has been added to the Fiber class to support the padding

## further implementation of eqpt_config json parameters:
* channel power spectrum input in ['SI']['power_dbm'] 
* power sweep input in ['SI']['power_range_db'] : [delta_min, delta_max, step] from the power_dbm value
           => loop from power_dbm+delta_min to power_dbm+delta_max in step
           => for example [-1,0,0.5] will simulate power_dbm-1 and power_dbm-0.5
* span padding is now read and implemented from eqpt_config[Spans][padding]
* con_in, con_out values are now read from eqpt_config if con_in/out are not defined in the network topology or values to None
* EOL span loss ageing parameter is now read from eqpt_config . EOL span loss margin is added to con_out in Fiber class (no specific attribute)
* span_max length default value is now read from eqpt_config in order to automatically split long fiber spans > span_max_length.
* update convert.py parser to provide None default values so that the eqpt_config json is the reference input for default values (don't have multiple default value sources implemented)

## code improvement and bug fixes
* improvement of the calculated length code to split long fiber > span_max_length. + it now takes into account the padding to set the min span length after splitting.
* fibx bug in setting the ROADM egress reference power in power mode (caused by previous code merge)
* remove coronet network outdated files that would cause the code to crash if they were used
* Improve errors handling so that an error message is displayed and the code exit cleanly. 
* Catch equipment configuration errors (for example input a fiber or an amplifier type_variety in the network topology that is not defined in the eqpt_config) => the error message clearly indicates that an eqpt definition is missing or mispelled instead of crashing for no obvious reason.
* rename all connector_loss_in/out with con_in/out

## update unitary tests
* update the tests so they take into account the code changes and can complete properly
* add a unitary test for the new fixed gain edfa to check its NF calculation


jeanluc-auge:
Thanks, 
I would suggest to do the fixes on the top of the 2nd PR to make sure that they are not competing with the latest change proposals

PR #137, opened 2018-10-26
Path disjunction

EstherLerouzic:
This pull request contains
- a new feature on path disjunction
- a change in power/nb_channel parameters from mandatory to optional, with a safe check on consistancy between spacing and baudrate
- a bug correction in route constraint import


examples/path_requests_run.py
```diff
@@ -46,7 +47,8 @@ def requests_from_json(json_data,equipment):
     requests_list = []
 
     for req in json_data['path-request']:
-        #print(f'{req}')
+        # print(f'{req}')
```
dutc: Remove this dead-code?

```suggestion
```

examples/path_requests_run.py
```diff
@@ -59,15 +61,52 @@ def requests_from_json(json_data,equipment):
...
+            fmax = trx_params['frequency']['max']
+            max_recommanded_nb_channels = automatic_nch(fmin,fmax,
+                min_recommanded_spacing)
+                
+            if req['path-constraints']['te-bandwidth']['max-nb-of-channel'] <= max_recommanded_nb_channels :
+                params['nb_channel'] = req['path-constraints']['te-bandwidth']['max-nb-of-channel']
+            else:
+                temp = params['baud_rate']
 
+                msg = f'Requested channel number is not consistant with frequency range  : \
```
dutc: Something like this might be easier to read (using `textwrap.dedent`)…

```suggestion
                msg = dedent(f'''
                Requested channel number is not consistent with frequency range:
                {fmin*1e-12} THz, {fmax*1e-12} THz and baud rate: {temp*1e-9} GHz
                min recommanded spacing is {min_recommanded_spacing}
                max recommanded nb of channels is {max_recommanded_nb_channels}
                Computation stopped.''')
```

gnpy/core/convert.py
```diff
@@ -155,7 +155,7 @@ def convert_file(input_filename, filter_region=[]):
                                         'longitude': x.longitude}},
               'type': 'Fused'}
              for x in nodes_by_city.values() if x.node_type.lower() == 'fused'] +
-            [{'uid': f'fiber ({x.from_city} → {x.to_city})-{x.east_cable}',
+            [{'uid': f'fiber ({x.from_city} \u2192 {x.to_city})-{x.east_cable}',
```
dutc: I don't think this file needs to be changed.

tests/data/excelTestFile_services_expected.json
```diff
@@ -53,9 +53,9 @@
       }
     }
   ],
-  "synchronisation": [
+  "synchronization": [
```
dutc: We should document a spelling guide so that we are consistent w/r/t American-vs-UK spellings.

(I don't care which we prefer.)

tests/test_disjunction.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
```
dutc: Thank you for including a new test with this!

gnpy/core/convert.py
EstherLerouzic: OK for me as long as the generated uid can be used in excel sheet route constraint. I checked that it was working with this code before you made the utf8 modif...

tests/data/excelTestFile_services_expected.json
EstherLerouzic: Sure!

tests/test_disjunction.py
EstherLerouzic: you are welcome :)

gnpy/core/convert.py
dutc: We can change the identifiers if we'd like, but one advantage of keeping this Unicode character is that it will suss out other internationalisation issues. 

We don't want everything to work smoothly for "Lyon" then blow up on "Rhône"!

EstherLerouzic: OK !

EstherLerouzic:
I have just done the resquested changes + removed the lines forgotten when adding dedent message

gnpy/core/service_sheet.py
```diff
@@ -105,7 +111,9 @@ def __init__(self,Request,eqpt_filename):
             # print(msg)
 
         self.loose = 'loose'
+        print(Request.is_loose)
```
dutc: What are the ranges of values? 

```
Request.is_loose in {'loose', 'no'} # what other possibilities?
```

gnpy/core/service_sheet.py
```diff
@@ -73,8 +73,14 @@ def __init__(self,Request,eqpt_filename):
             exit()
         # excel input are in GHz and dBm
         self.spacing = Request.spacing * 1e9
-        self.power =  db2lin(Request.power) * 1e-3
-        self.nb_channel = int(Request.nb_channel)
+        if Request.power is not None:
```
dutc: Can we switch this to `request` with a lowercase `r`?

dutc: Nevermind. We'll fix this and a couple other superficial things with a follow-up PR.

dutc:
Looks good over all.

We should submit a small clean-up PR after to fix some superficial things.

gnpy/core/service_sheet.py
EstherLerouzic: In fact I had in mind **'loose' and 'strict',** but I think I mixed up a bit with the excel format. right now the excel sheet should have a yes or no in 'routing: is loose ?' column, but the code uses loose and strict. Moreover, somebody asked me for a stricter constraint, which would be 'use exactly the succession of node' as a path, so there might be an 'exact' term in a near future if this proves to be useful.

PR #141, opened 2018-10-28
Simplify automatic spacing

ojnas:
Simplification of automatic_spacing function. It should do the same thing.

### DIFF
--- a/gnpy/core/equipment.py
+++ b/gnpy/core/equipment.py
@@ -182,13 +182,7 @@ def automatic_spacing(baud_rate):
     """return the min possible channel spacing for a given baud rate"""
     spacing_list = [(38e9,50e9), (67e9,75e9), (92e9,100e9)] #list of possible tuples
                                                 #[(max_baud_rate, spacing_for_this_baud_rate)]
-    acceptable_spacing_list = list(filter(lambda x : x[0]>baud_rate, spacing_list))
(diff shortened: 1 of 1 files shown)


gnpy/core/equipment.py
```diff
@@ -182,13 +182,7 @@ def automatic_spacing(baud_rate):
...
     spacing_list = [(38e9,50e9), (67e9,75e9), (92e9,100e9)] #list of possible tuples
                                                 #[(max_baud_rate, spacing_for_this_baud_rate)]
-    acceptable_spacing_list = list(filter(lambda x : x[0]>baud_rate, spacing_list))
-    if len(acceptable_spacing_list) < 1:
-        #can't find an adequate spacing from the list, so default to:
-        return baud_rate*1.2
-    else:
-        #chose the lowest possible spacing
-        return min(acceptable_spacing_list, key=itemgetter(0))[1]
+    return next((s[1] for s in spacing_list if s[0] > baud_rate), baud_rate*1.2)
```
dutc: Would this be closer to the original?

```suggestion
    return min((s[1] for s in spacing_list if s[0] > baud_rate), default=baud_rate*1.2)
```

ojnas:
I think my version should always return the same result as your suggestion as long as `spacing_list` is sorted. Also it should be more efficient since it does not have to go through the whole list.

Issue #130, opened 2018-10-22
Excel file example and source/destination


ojnas:
Looking at CORONET_Global_Topology.xls, it might be expected that any node from the "Nodes" sheet could be a source or destination node. But running the following

`python transmission_main_example.py CORONET_Global_Topology.xls Abilene Albany`

results in the confusing output

```
invalid souce node specified, did you mean:                  
[]?                  
'Abilene', replaced with trx Albuquerque
invalid destination node specified, did you mean:                
[]?                
'Albany', replaced with trx Atlanta
```

I assume this is because both Abilene and Albany are degree-2 nodes, which according to Excel_userguide.rst means their type will be set to ILA, and source/destination must be of type ROADM. Maybe it would be less confusing to just set all nodes to ROADM if type is not specified?

dutc:
We can discuss this on our next coder's call. @jeanluc-auge @EstherLerouzic, do you have any comment on this issue?

ojnas:
Hi @EstherLerouzic !

> This was our choice because in usual topologies, ILAs (or Fused) come more often than ROADM and because it leaves the possibility to choose a degree-2 node to be either ILA or Fused which may be interesting to optimize the design.

Yes, but on the other hand, in many topologies most ROADM nodes are only degree-2 (in CORONET Global roughly half the nodes are degree-2). Whatever you choose as default, you will have to set the type explicitly for many nodes.

In general I think defaults should be simple and not try to make "intelligent" guesses since this will just be confusing if you don't know exactly the algorithm behind the guessing. In this specific case, if you don't read Excel_userguide.rst (yeah, I know you should) or look at the code you will be confused about why half the nodes in the xls are not present in the json topology file. Also, I think it would probably be better to return an error message and then stop if the user provides source or destination nodes without transponders instead of running with source and destination nodes in completely different locations.

The choice between ILA and Fused is not automatic if I understand correctly.

Anyway, I'm fine if you want to keep the default as is but then something should be done about the CORONET Global example (e.g. filling the type column explicitly like you say) since that is the first example most users will see.

EstherLerouzic:
Hi @ojnas !
I understand your point ! and you are right I thought the fused choise was automatic at one point but it is not, sorry !
Let's talk of that during our next conf call ! Any of your solution (keeping or changing optional fields, correcting file, and removing the "too smart decision) make sense but I would like to know what is the preference of the group 
thanks a lot for contributing ! I really appreciate your inputs !

[...]

PR #143, opened 2018-10-30
Explicitly set all nodes in CORONET example to ROADM

ojnas:
See #130

Also updated README to reflect that by default propagation is now between Abilene and Albany again.

Will propose a separate PR to avoid crashing because source and destination are the same.


dutc:
References #130 

Since the Coronet files are generated from the original .XLS, we may want to fix the converter.

Otherwise, we risk these changes being accidentally reverted in future work.

Also, could you separate commit 48198bd from this PR?

dutc:
My mistake. I saw the change to the .json but I didn't see the change to the .xls. You did the right thing.

Don't worry about separating your commits for this PR. I'll handle it on my end.

ojnas:
Ok, thanks, but I would still like to understand how to separate commits. I wanted to do it but gave up after fighting with git for too long...

dutc:
We've previously discussed doing a `git` warmup call for the code contributors to cover some of these types of issues. If you'd be interested, I'll reraise this as something we might want to do.

In the meantime, check out: https://github.com/k88hudson/git-flight-rules

dutc:
We've previously discussed doing a `git` warmup call for the code contributors to cover some of these types of issues. If you'd be interested, I'll reraise this as something we might want to do.

In the meantime, check out: https://github.com/k88hudson/git-flight-rules

PR #144, opened 2018-10-30
Change source and destination logic in transmission_main_example.py

ojnas:
See #130.

- if provided destination is the same as source, replace the destination to avoid crash

- fix the bug that replaced destination with same node as source

- improve logic to check both source and destination for exact match first, then partial match

- move printing of messages about replacing or picking source/destination node to the end to avoid confusion


ojnas:
Since this PR has been open for a month and it was approved two weeks ago, is there anything more needed to merge the patch?

Meanwhile, the bug mentioned [here](https://github.com/Telecominfraproject/oopt-gnpy/issues/130#issuecomment-433989006) is still present both in master and develop even though issue #130 was closed.

jeanluc-auge:
I approve this change
I have read and fetch the code to test it

Issue #147, opened 2018-11-08
power sweep and default power_range_db


ojnas:
I don't know if the following is a bug or intended behavior but to me it's slightly confusing. In the transmission main example there is this code:
```
power_range = list(arange(*equipment['SI']['default'].power_range_db))
        last = equipment['SI']['default'].power_range_db[-2]
        if len(power_range) == 0 : #bad input that will lead to no simulation
            power_range = [0] #better than an error message
        else:
            power_range.append(last)
```
In the default `eqpt_config.json` there is `"power_range_db": [0,0,0.5]`. With this input `power_range` becomes an empty list. It is confusing to have the default input being referred to as "bad input" in the code. Also, setting ` power_range = [0]` in this case may lead to unexpected behavior. For example, if a user inputs `"power_range_db": [1,1,0.5]` it may be expected that `power_range` becomes 1 and not 0.

ojnas:
Also, the use of `numpy.arange` may have unexpected results because of floating point error. For example:
```
>>> arange(1, 1.1, 0.1)
array([1. , 1.1])
```
In this case, since the last value is appended with `power_range.append(last)`, the value 1.1 will be included twice.

dutc:
We'll look into this this week.

jeanluc-auge:
basically tjhe arange function was not doing what I needed, so I tried to modify it.
Knowing that the result of the power range is providing the power excursion around the power_dbm setting (the parameter in the line just before), feel free to propose an enhancement.
Thanks

ojnas:
I would propose using `numpy.linspace` instead. I will submit a PR and then you can check if it does what you want.

dutc:
In general, I think using `numpy.arange` is better than using `numpy.linspace`.

I think these values should be specified as open-closed intervals `[start, stop)` which is consistent with `numpy.arange`, `range`, and slice notation.

Currently, we use the default `json.JSONDecoder` when we call `json.loads`. This comes with a set of assumptions that are causing the above example to break. I think we can start adding some sophistication to our decoding. As a first step, we can use `decimal.Decimal` to preserve user intention and avoid issues with imprecise encoding of real numbers with IEEE-754 binary fp. It's likely that we may have other floating point decoding issues that this can help address, and we can also use this as an opportunity to reorganize some complexity into the decoding/parsing logic.

I think the right answer looks something like:

```python
from json import loads
from decimal import Decimal
from numpy import arange

raw_input = '''
[
[...]

PR #156, opened 2018-11-14
improve power range handling - revised version

ojnas:
See #147.


jeanluc-auge:
I approve this change
I have fetched and tested the code, it works fine

PR #154, opened 2018-11-13
Service aggregation clean

EstherLerouzic:
For you to see my work on
- aggregation feature
- automatic mode selection feature
- cost feature
+ various small fixes

Esther

PR #166, opened 2018-11-19
Roadm fix

jeanluc-auge:
Please first merge @EstherLerouzic PR #154 before this one


ojnas:
Here you changed the Open ROADM noise figure definition (compared to #154):

`nf_avg = pin_ch - polyval(self.params.nf_model.nf_coef, pin_ch) + 58`

I guess this means that the `nf_coef` in eqpt_config.json now actually defines incremental OSNR (and therefore indirectly noise figure) vs. Pin. In this case the name `nf_coef` may be a bit confusing but this is a minor detail.

jeanluc-auge:
Yes, at some point we hesitated between NF coeff and incremental OSNR but finally decided to keep the incremental OSNR, albeit now as a function of Pin. 
Yes it is confusing, I will change it in another PR.

jeanluc-auge:
I have update the PR because add channels did not have the same behaviour as express channels.
So I update the loss to effective loss when considering the ingress power (pref.pi), which is not the same for xpress vs add channels.

jeanluc-auge:
sorry for the mess
a reset --hard later (...) I'm back to the initial ROADM power management. What I did does not work with the power sweep. So I revert the initial behaviour albeit with some improvement that correct a bug when the ingress amplifier was saturating.

jeanluc-auge:
closing it to clean up and cherry-pick my commits on the new develop branch

PR #175, opened 2018-11-30
Few minor details and typos in README

MiquelGA:
Kindly, few minor details modified. I also mention: Additional details are provided by doing ``transmission_main_example.py -h``.


ojnas:
The change in the README to "Albuquerque and Atlanta" is not consistent with the `transmission_main_example.py` code currently in the develop branch, which actually propagates between Abilene and Albany. The script in the master branch propagates between Albuquerque and Atlanta so there the README is indeed wrong. But if you would merge both changes from develop to master now it would be wrong again.

Issue #116, opened 2018-10-04
Transceiver Tx OSNR


ojnas:
It seems there is currently no simple way to set the output OSNR from a source transceiver. The `transmission_main_example.py` script calls the `propagate` function in core module `request.py`, which in turn calls the `create_input_spectral_information` function in `info.py` where the ASE noise is "hard-coded" to zero. So the Tx OSNR is always infinite which does not seem very realistic (e.g. the Voyager transponder specifies Tx OSNR to 40 dB but I have seen other specifications as low as 32 dB).

ojnas:
In addition, Tx output in-band and out-of-band OSNR is sometimes specified separately, see e.g. table 3 in this datasheet:

[https://www.cisco.com/c/en/us/products/collateral/routers/network-convergence-system-5500-series/datasheet-c78-741079.html](url)

I think the existence of out-of-band ASE noise also means that the current ROADM model in `elements.py` is a bit simplistic but I will post a separate issue about that...

dutc:
Thank you for this feedback! @jeanluc-auge is putting together some comments and will provide feedback after the TIP Summit.

jeanluc-auge:
Thank you,
this is a very fair point that we should address indeed.
By the way, we should also consider ROADM amplified Add & drop (A&D) contributions. For example there could be a library of ROADM types with different OSNR contributions. Non amplified A&D ROADM architectures would have a default 100dB OSNR, while amplified A&D contribution could be defined (typically for 1st gen MCS architectures that have high losses and Edfa arrays).
Starting wiht the Tx OSNR as you suggest, it could be a good way to start contributing to the code: would you like to take this up? Of course we can help you getting started,
cheers

ojnas:
@jeanluc-auge, I thought a little bit about how to implement Tx OSNR. One problem is how to handle out-of-band ASE. This is a transceiver parameter but the penalty depends on the ROADM add/drop-structure (filtered vs. non-filtered). The `SpectralInformation` tuple includes signal and (in-band) ASE power per channel. So what to do when network elements just propagate `SpectralInformation` independently?

[...]

PR #177, opened 2018-12-11
Add Tx OSNR

ojnas:
This adds support for a Tx OSNR parameter, see #116. ASE noise out from the transponder is set in the `create_input_spectral_information` function according to this parameter.

The Tx OSNR parameter is defined in `eqpt_config.json`. Currently it is set to a high value, 45 dB, both in SI default and in all trx modes.

So far, no distinction is made between in-band and out-of-band Tx OSNR. Penalty due to ASE noise from other transponders and filterless add/drop is not taken into account.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -253,6 +253,8 @@ The modes are defined as follows:
 +----------------------+-----------+-----------------------------------------+
 | `roll_off`           | (number)  | Not used.                               |
 +----------------------+-----------+-----------------------------------------+
+| `tx_osnr`            | (number)  | In dB. OSNR out from transponder.       |
++----------------------+-----------+-----------------------------------------+
 
 Simulation parameters are defined as follows.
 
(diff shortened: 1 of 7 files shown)


jeanluc-auge:
I approve his change
I have read and fetch the code to test it

EstherLerouzic:
seems that this new features crashes when mode is not filled
I propose a correction

HidekiNishizawa:
The parameter Tx OSNR depend on each transceiver, so it should be put from the transceiver respectively in the transponder. 
Reference:
Open ROADM defined Minimum transmitter OSNR = 33dB
http://openroadm.org/download.html
Cablelabs defined that the transceiver must provide a transmitter OSNR of 35 dB or higher.
https://specification-search.cablelabs.com/P2PCO-SP-PHYV2.0
These organizations tend to relax optical parameter's spec for cost reduction, lower the barrier for new comers.

Issue #179, opened 2018-12-12
Programs crashes when NLI = 0


ojnas:
Both `transmission_main_example.py` and `path_requests_run.py` crashes when there is no nonlinear interference noise (NLI = 0) with the following error:

`TypeError: unsupported operand type(s) for /: 'NoneType' and 'int'`

This can be triggered e.g. by setting gamma = 0 under fiber parameters in eqpt_config.json. The reason is that the `_calc_snr` function in `class Transceiver` does not calculate OSNR_NLI when NLI is lower than 1e-20 (leaving it as None), which results in an error when calculating the average OSNR_NLI in the main script.

(This issue is related to #178.)

jeanluc-auge:
-bug is fixed, I simulated 2 transponders connected back to back and it returns inf SNR
-I checked gamma = 0, same result
I propose to close this issue.

dutc:
I also checked this issue with zero gamma and did not see the bug.

I ran:
`$ python transmission_main_example.py CORONET_Global_Topology.json`
and
`$ python transmission_main_example.py CORONOT_Global_Topology.json -e eqpt_config.json`

My `eqpt_config.json` changes:
```
      "Fiber":[{
            "type_variety": "SSMF",
            "dispersion": 1.67e-05,
            "gamma": 0
            },
            {
            "type_variety": "NZDF",
            "dispersion": 0.5e-05,
            "gamma": 0
            },
            {
            "type_variety": "LOF",
            "dispersion": 2.2e-05,
            "gamma": 0
            }
```

ggrammel:
confirm James' findings. can be closed

PR #180, opened 2018-12-13
Fix OSNR calculation when ASE or NLI is zero

ojnas:
Proposed fix for #179 using `numpy.divide` to handle division by zero (resulting in `inf`) in OSNR calculations.


jeanluc-auge:
I approve this change
Good work!
it no longer crash when the same source and destination is entered, instead it now returns inf
I have read and fetched the code to test it

Issue #178, opened 2018-12-12
OSNR calculation in Transceiver class for a list of services


ojnas:
When running `path_requests_run.py` with a service file specifying a list of services, OSNR calculation will be wrong under the following circumstances:

1. ASE or NLI is lower than 1e-20, see code in `_calc_snr` function in `class Transceiver` (this could for example happen if all elements are linear, in which case NLI = 0) and
2. OSNR has already been calculated for a different service going to the same destination transceiver.

In this case, if e.g. NLI = 0, `self.osnr_nli` and `self.snr` will not be updated. Instead, values for the previously computed path going to the same destination will be reported.

(If there was no previous path and NLI = 0, the program will instead crash which also is not ideal.)

ggrammel:
seems not fixed. probably linked to #179 which had a regression too

changed NLI in eqpt_config.json to 0.0

ran
```
python path_requests_run.py meshTopologyToy.xls meshTopologyToy.xls
```
Result:
```
Computing path from trx R_A to trx R_C
with path constraint: ['trx R_A', 'R_B', 'trx R_C']
Computed path (roadms):['roadm R_A', 'roadm R', 'roadm R_B', 'roadm R_C']

Traceback (most recent call last):
  File "path_requests_run.py", line 144, in <module>
    line = [f'{pths[i].source} to {pths[i].destination} : ', f'{round(mean(p[-1].snr),2)}',\
  File "/anaconda3/lib/python3.6/site-packages/numpy/core/fromnumeric.py", line 2957, in mean
    out=out, **kwargs)
  File "/anaconda3/lib/python3.6/site-packages/numpy/core/_methods.py", line 82, in _mean
    ret = ret / rcount
TypeError: unsupported operand type(s) for /: 'NoneType' and 'int'
```

ggrammel:
tried the following with and without setting gamma to 0.0 result was the same in both cases

```
>> python path_requests_run.py meshTopologyToy.xls meshTopologyToy.xls
Computing path requests meshTopologyToy.xls into JSON format
CRITICAL: missing _east_ header above other headers (hierarchical) _ ABORT
```
Then repeated the test with
```python path_requests_run.py meshTopologyExampleV2.xls meshTopologyExampleV2.xls -o tototo```
and this works. if ```meshTopologyToy.xls``` doesn't work, we could either fix or remove before committing to master

dutc:
Can we get @EstherLerouzic to comment on these two files?

[...]

PR #182, opened 2018-12-17
Corrections

EstherLerouzic:
Several fixes handling some bugs

- output file of -o option in path_requests_run.py was not correctly
  handled in case of path indirection eg ../../bar.foo
- fiber constraint is not correctly handled in case of several parrallel directions
  of same fiber name. remove if from the set of constraint till problem is not solved
  This will need additional work to be compliant with fiber constraint
- loose_list was not correctly indexed in request.py. assume the first element attribute (except
  of the transceiver) applies for the whole list. 
  This is a fast patch and will need further corrections to handle a detail per node loose definition
- handle the case when path.snr is none in the optimization of mode process
  maybe addresses part of #178 
- a cutoff value needed when networks have a great number of nodes and links (tested on ~ 500)


EstherLerouzic:
I will rebase my work on the novel develop when bug on tx_osnr is merged on develop

PR #183, opened 2018-12-17
Fixes concerning service sheet reading

EstherLerouzic:
Fixes concerning service sheet reading
    
    - indicate which request has incorrect tsp type or mode
    - correct wrong reading from excel : if string values contain only a number
      xlrd interprets as a float eg 1 -> 1.0  . A function is used to correct this
      when filling the Request class


EstherLerouzic:
I will rebase my work on the novel develop when bug on tx_osnr is corrected

Issue #185, opened 2018-12-19
novel update with tx_osnr feature fails with automatic mode selection


EstherLerouzic:
**Describe the bug**
when no mode is indicated, the automatic mode selection fails
python path_requests_run.py meshTopologyExampleV2.xls meshTopologyExampleV2.xls -o tototo 
Traceback (most recent call last):
  File "path_requests_run.py", line 307, in <module>
    rqs = requests_from_json(data, equipment)
  File "path_requests_run.py", line 111, in requests_from_json
    requests_list.append(Path_request(**params))
  File "/home/user/workspace/TIP/gnpy/gnpy/core/request.py", line 42, in __init__
    params = RequestParams(**params)
TypeError: __new__() missing 1 required positional argument: 'tx_osnr'

I will propose a correction today + tests to avoid future problems.

jeanluc-auge:
I have an alternate proposal:
-this slows the code down
-the path calculation is no longer transponder agnostic
-we will have the same issue with ROADM noise contribution
Instead I propose to add the transponder Rx OSNR (and ROADM add/drop by the way) contribution at the end of the calculation with just an OSNR sum. I have checked, it gives the same result. This small ase contribution will not make any difference in the nli calculation anyway, so it doesn't matter if we had it at the end of the path. This way it can be added when chosing the transponder and you don't need ot re-run the simulation for every transponder mode.

ojnas:
@jeanluc-auge 

Your proposal to add tx_osnr contribution at the end in order to speed up calculation is what I had in mind in my comment here: https://github.com/Telecominfraproject/oopt-gnpy/pull/186#issuecomment-448991213.

On the other hand I don't really agree with this argument:
>-the path calculation is no longer transponder agnostic

The path calculation already depends on baud_rate (and it may also depend on roll_off when Raman is included) so it's not really transponder agnostic anyway.

I proposed an alternative way to speed up calculation by caching results in my comment here: https://github.com/Telecominfraproject/oopt-gnpy/pull/186#issuecomment-453526197. I don't have a strong opinion on which of the proposals is better but somehow it feels less "future proof" to add noise contributions at the end. Even if it gives the same result now there is no guarantee that this will be the case for new network element models.

ggrammel:
```
python path_requests_run.py meshTopologyExampleV2.xls meshTopologyExampleV2.xls -o tototo
```
creates the following output:
```
N channels =  95
N channels =  95
N channels =  95
[...]

PR #186, opened 2018-12-19
bug fix due to tx_osnr add

EstherLerouzic:
This is a correction to address bug pointed in  #185


ojnas:
Thanks, I did not test the tx_osnr and automatic mode selection features together.

Now I looked closer at the code in the `propagate_and_optimize_mode` function. It looks like the simple addition of tx_osnr means you now have to do a full propagation for every mode whereas before it was enough to do it once for every baudrate? I wonder if it would be possible to make it more efficient by adding the tx ase noise separately.

Another thing: you check that the mean OSNR for all channels is above threshold:
```
if round(mean(path[-1].snr+lin2db(b/(12.5e9))),2) > m['OSNR'] :
                        found_a_feasible_mode = True
```
Wouldn't it make more sense to check that the minimum OSNR over all channels is above threshold?

EstherLerouzic:
Indeed, if the tsp has many modes and the path has low snr, it lengthen the computation time. I have tested on a large network (500 nodes) with >100 demands (using the automatic mode selection with only two modes) the computation time was about 270s before the change and 340s after, meaning that on this particular case it was lengthen by a factor of 25%. With more modes this may increase even more.
If you can find a way to avoid propagation on each mode, this could be more efficient :)

agreed with your last point : min snr instead of mean snr makes sense

ojnas:
A simple way to make it more efficient would be to at least avoid propagation for each mode when the tx_osnr is the same. Here is a proposed implementation (on top of your tx_osnr_bug_correction branch) where I cache calculated snr in a dict using tx_osnr as key (I'm not sure using a float as key is a good idea but it seems to work):

https://github.com/ojnas/oopt-gnpy/tree/tx_osnr-mode_selection-efficiency

Could you test this branch on your large network to see how it impacts computation time and whether results are the same?

Issue #190, opened 2019-01-09
coords property of class node type error on return value


davidBoertjes:
**Describe the bug**
When asking for the value of the *coords* property of a *node* object the return value has a type error.

**To Reproduce**
in iPython, run the provided example:
```>>>runfile('transmission_main_example.py')```
then type
```
>>>path[2].coords
Traceback (most recent call last):

  File "<ipython-input-17-0b3fcebcaf53>", line 1, in <module>
    path[2].coords

  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64\envs\gnpy\lib\site-packages\gnpy-0.1.3-py3.6.egg\gnpy\core\node.py", line 39, in coords
    return tuple(self.lng, self.lat)

TypeError: tuple() takes at most 1 argument (2 given)
```

**Expected behavior**
expect a tuple of the coordinates of latitude and longitude:
'''
>>>path[2].coords
 (0, 2)
'''

**Environment:**
 - OS: Windows
 - Python version 3.6.8 |Anaconda, Inc.| (default, Dec 30 2018, 18:50:55) [MSC v.1915 64 bit (AMD64)]

**Suggested solution**
Change the following line (line 39 of node.py):
`        return tuple(self.lng, self.lat)`
to
`        return (self.lng, self.lat)`

PR #191, opened 2019-01-10
fix for issue #190

davidBoertjes:
Change to node.py to fix return value error in *coords* property which was intended to build a tuple from two values for latitude and longitude.


davidBoertjes:
I apologize for the trivial PR.  This is my first attempt in this repo.  Any future ones from me will be larger.

ggrammel:
David,

No need to apologize, we appreciate any comment.

Gert

PR #197, opened 2019-01-22
speed up automatic mode selection by caching calculated snr

ojnas:
Proposal to avoid slowdown of automatic mode selection caused by introduction of tx_osnr and therefore having to perform propagation for each mode.  This is accomplished by caching calculated snr so that the already calculated value can be reused if tx_osnr is the same for multiple modes.

See discussion here: PR #186 

@EstherLerouzic, it would be great if you could test calculation again on your large network mentioned [here](https://github.com/Telecominfraproject/oopt-gnpy/pull/186#issuecomment-449382239) with and without this PR to see 1) impact on computation time and 2) that the results are the same.

### DIFF
--- a/gnpy/core/request.py
+++ b/gnpy/core/request.py
@@ -416,18 +416,23 @@ def propagate_and_optimize_mode(path, req, equipment, show=False):
             # print(modes_to_explore)
             # step2 : computes propagation for each baudrate: stop and select the first that passes
             found_a_feasible_mode = False
+            snr_cache = {}
             # TODO : the case of roll of is not included: for now use SI one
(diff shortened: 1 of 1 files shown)


dutc:
Thanks for this submission @ojnas. 

We'll review it on today's Coder's Call, and determine if we'll slate it for inclusion in this week's v1.1 release or postpone it until after.

ojnas:
@jeanluc-auge, ok, no problem, I'm fine with adding tx_osnr at the end as well. I will have a look at your PR and maybe comment on it.

jeanluc-auge:
Ok great, thanks Jonas

PR #198, opened 2019-01-22
Sys magins

jeanluc-auge:
Tx_OSNR added at the end of propagation
Terminal ROADM add drop OSNR contribution (in eqpt_config / ROADM)
System margins contribution (in eqpt_config / SI)
code cleaning for the SI function


jeanluc-auge:
I see a bug: I will correct and push a new version

PR #199, opened 2019-01-23
Sys margins

jeanluc-auge:
Tx_OSNR added at the end of propagation
Terminal ROADM add drop OSNR contribution (in eqpt_config / ROADM)
System margins contribution (in eqpt_config / SI)
code cleaning for the SI function

### DIFF
--- a/examples/eqpt_config.json
+++ b/examples/eqpt_config.json
@@ -109,7 +109,8 @@
             "power_dbm": 0,
             "power_range_db": [0,0,0.5],
             "roll_off": 0.15,
-            "tx_osnr": 40
+            "tx_osnr": 40,
+            "sys_margins": 0
             }],
       "Transceiver":[
             {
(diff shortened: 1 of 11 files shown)


ojnas:
I had a quick look and just have a couple of minor comments:

-It's a bit confusing to see assignments like `snr_added += db2lin(-s)`, where the right-hand side is actually the inverse of an snr. Maybe this could be rewritten in a clearer way.

-I'm not sure it makes sense to add tx_osnr and add_drop_osnr to osnr_nli. I think osnr_nli should only include nonlinear interference noise computed by the GN-model.. 

-The README should probably be updated to explain the new parameters add_drop_osnr and sys_margins

jeanluc-auge:
Hi jonas, you are absolutely right, it makes no sense to add it to the osnr_nli: I must have been in automatic mode ;) I will correct this.
For the snr_added feel free to propose, 
and for the readme we agreed during the Tuesday call that it should indeed be our next task
Thanks a lot for checking and suggestions

Issue #209, opened 2019-01-30
Questionable auto-correction in transmission_main_example.py


ggrammel:
Is this really the intended behaviour?

```
python transmission_main_example.py CORONET_Global_Topology.xls z z

<snip>

(Invalid source node 'z' replaced with trx Zurich)
(Invalid destination node 'z' replaced with trx Abilene)
```

I get that one would like to be tolerant against typos, but this goes a bit too far.
- My preference is to simply throw an error message if the two Transponders do not exist
-    In case the team still want to maintain the auto-correction, there should be an option --strict to ensure consistency when desired
-    even in the auto-correction case I would suggest to ensure that at least the initial letter of the location is the same. Resolving 'z' to Zurich looks OK, but resolving the second 'z' to Abilene is a bit much.

ojnas:
This was discussed before in #130, see this [comment](https://github.com/Telecominfraproject/oopt-gnpy/issues/130#issuecomment-432563556) and comments below it. As you can see in that discussion, I tend to agree with you that the current auto-correction goes a bit to far. A `--strict` option seems like a good idea.

PR #212, opened 2019-02-04
--strict option for source and destination nodes

ojnas:
This implements a `--strict` option for matching source and destination nodes in transmission_main_example.py as proposed in #209.

Usage:
when the option is not supplied, the script works as before
supply -s or --strict for allowing partial matching
supply -ss or --strict --strict for only allowing full matching


dutc:
The strict mode should be the default, and we should fix our input data to include the actual, human-printed city name in the metadata.

`transmission_main_example.py` should allow the user to specify the pretty-printed city name (which selects the transceiver/`"trx_" node.) We can add a `--fuzzy` or similar flag to match fuzzily using a simple heuristic like Levenshtein.

Currently, fuzzy matching is basically always used, because the transceiver nodes are specified as `trx_CITY` with the CITY name transformed by replacing spaces with underscores, etc.

jktjkt:

PR #215, opened 2019-02-19
Roadm equalization

jeanluc-auge:
major refactor:
- auto-design improvement for amplifier selection
- new dual stage amplis: replace hybrid raman implem
- read operational power settings
    gain settings are still read in gain mode (power_mode = False in eqpt_config.json)
- roadm architecture change: target_pch_out_db sets the roadm output power (before booster/egress amp) in gain mode and power mode. roadm_loss is suppressed.
- power_mode reads operational power settings or calculates them. Operational power settings defines a delta power wrto the nominal eqpt_config.json['SI'] power. New delta p column in Eqpt sheet for the xls parser
- in gain mode the power can no longer be set: it is the result of amplifier gain configuration. Gain mode should not be used in auto-design

PR #220, opened 2019-03-20
Api rest

ahmed-triki:
This is an implementation of an API rest for GNPy. The http server is based on flask, so the installation of this package is mandatory. The input/output of the api-rest are modeled by a yang data model.
In addition to the Rest mode, the new code keeps the possibility to use the current CLI (Command Line Interface) mode.
To activate the Rest mode, the user has to launch the following command : 
python path_requests_run.py --rest
The API-rest client has to take into account the following connection parameters: 
- URL: http://<ip_address>:5000/gnpy/api/v1.0/files 
- authentication: username: gnpy/ password: gnpy
- Rest-Method: POST / The body of the request is composed of the topology and the service data in json format.
This pull-request has to be merged after #215 .
Todo: add explanation of the body and the response models of the API-REST.


ahmed-triki:
Thank you @ojnas  for your interest to the code.
Exactly, the actual yang files are not used directly by the code, they are only included as a reference. However, the Post request has to be compliant with the yang model in order to be understandable by GNPy code. I have already used these yang models in T-PCE in order to enable T-PCE to interrogate GNPy and check the feasibility of the computed path.

ojnas:
Maybe this REST API functionality should be part of the gnpy core library, i.e. placed in a module in the gnpy/core folder and not just as a part of the `path_requests_run.py` example script as it is now?

dutc:
@ojnas I agree that the REST functionalit should be extracted from `path_requests_run.py` and we have discussed putting it into a separate library from http://github.com/telecominfraproject/oopt-gnpy

We discussed creating a repo like http://github.com/telecominfraproject/oopt-interfaces or some such to collect higher-level "interfaces" for `gnpy`, but we're still discussing if this is the right name. The reason to split this off into its own repo are:
- allows the REST API to target a specific version of the core, underlying library
- allows contributors to the core, underlying library (who may have no background in Flask/REST APIs) to make changes without worrying about breaking REST API
- keeps the dependency footprint of the core library small (since Flask may not be an approved or preferred tool within all organizations)

dutc:
Hi, @ahmed-triki, as per our call on Thursday, I have created http://github.com/telecominfraproject/oopt-tools for "tool"-level code for OOPT. This would allow us to organize ourselves as follows:
- github.com/telecominfraproject/oopt-raman: scientific-level code
- github.com/telecomifnraproject/oopt-gnpy: core library code
- github.com/telecominfraproject/oopt-tools: user-facing tools code (REST API, command-line tools, etc.)

EstherLerouzic:
please refer to the new PR #309   for this work

jktjkt:

PR #221, opened 2019-04-04
Interpol amp

jeanluc-auge:
* improvement of the advanced amplifier model with better interpolation between the advanced json description vs the spectrum information
* add a juniper amplifier example


jeanluc-auge:
## To be merged 2nd

PR #223, opened 2019-04-10
Add weight edge

EstherLerouzic:
Could you please have a look on this work ?
I have been requested several time to provide path computation on the length (km) basis instead of hop.
Here is what I propose to integrate this: 
I have added weigths on network edge in order to compute shortest path in length (km) and not in hop. 
- dijkstra_path with weight attribute
- sort all_simple_paths results as a function of paths' length instead of hop

Note that the feature only works with updated Networkx module. I have not made this change optional since in optical networks length is more relevant than hops.


EstherLerouzic:
I have just rebased onto develop. Now ready to merge

PR #244, opened 2019-05-27
correct test and example files with direct Trx reception

EstherLerouzic:
adding  -15 differential per target on receiver instead of 0
and adding attenuator in front of receivers in example networks
and test network, to keep same performances as before this change

### DIFF
--- a/examples/edfa_example_network.json
+++ b/examples/edfa_example_network.json
@@ -50,6 +50,21 @@
             }
         },
         {
+            "uid": "Att_B",
+            "type": "Fused",
+            "params":{
+                "loss":16
+            },
+            "metadata": {
+                "location": {
+                  "latitude": 2.0,
+                  "longitude": 0.0,
+                  "city": "B",
+                  "region": ""
+                }
+            }
+        },        
+        {
             "uid": "Site_B",
             "type": "Transceiver",
             "metadata": {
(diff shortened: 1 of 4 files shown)


ojnas:
Two questions:

- Why do you say "instead of 0"? If I understand the current code correctly the differential power (`dp`) depends on the loss of the following element. For a Transceiver the loss is 0 so the target would already be a negative differential, right?

- I don't think the current Transceiver code cares about the input power, i.e. it only calculates OSNR/SNR and does not check that input power is within some range, so why is this change necessary?

EstherLerouzic:
@ojnas thank you for your comment:
on question 1 :  the "instead of 0" referred to the roadm dp. Indeed the dp value depends on "node" loss and delta_power_range defined in eqpt_config.json, so it can be negative 0 or positive. In the case of a trx the loss is set to 0 (by the function span_loss). 
on question 2 : it's correct, the current transceiver does not care about its input power, performance is only limited by SNR. The purpose of this change was to put the finger on this fact, have more realistic power level on receiver and prepare future possible work on power limitation (not for the short term !).
I would prefer to integrate the change, but let's discuss this tomorrow within our weekly call

jktjkt:
I re-checked Codecov report. It complains because the newly added code path which sets `Transponder`'s RX power is actually not hit by any test case. So the complaint is reasonable because this pull requests adds code which is not tested.

I'll add a test case which exercises this path.

jktjkt:

Issue #118, opened 2018-10-12
EDFA NF from polynomial fit


ojnas:
In the `_calc_nf` method in the EDFA class, for the advanced model, the average NF is evaluated from target gain and polynomial coefficients like this:

`nf_avg = polyval(self.params.nf_fit_coeff, -dg)`

where `dg` should be the difference between target gain and (average) design-flat gain but is calculated like this:

`dg = max(self.params.gain_flatmax - gain_target, 0)`

So for target gain > design-flat gain `dg` will be 0 and average NF will be constant. It this really a correct model? I think the polynomial fit should be valid for any target gain and `dg` should simply be calculated as:

`dg = self.params.gain_flatmax - gain_target`

dutc:
Thank you for this feedback! @jeanluc-auge is putting together some comments and will provide feedback after the TIP Summit.

jeanluc-auge:
Hi,
thank you for your comments and taking the time to check the code.
This is not a bug, and I stand by this implementation.
* If gain < GAIN_FLATMAX, the nf polynomial will indeed follow gain variations
* but if gain > GAIN_FLATMAX, you enter the so called extended gain range where the internal VOA is at its min value (so little NF improvement) and the amplifier starts to tilt. The range of this extended gain region is not fully limited in the code (partially in auto-design) and in real use, whereas the validity range of any polynomial fit is. Typically most polynomial fits may only be validated within the amplifier gain range and are not safe to use outside (strange rebounds may happen with high polynomial orders). As already mentionned: considering the very limited NF variation in this region, it was deemed a better choice to fix the NF to its minimum value.

ojnas:
@jeanluc-auge, thinking a bit more about this, I have larger issues with the EDFA model. Your reasoning above applies to variable gain EDFAs (with internal VOA). I think an "advanced model" should be able to model any type of EDFA. For example, fixed gain EDFAs (without internal VOA) typically specifies a gain range as well as an optimal flat gain between gain_max and gain_min. So three gain parameters would be needed instead of the two (gain_min and gain_flatmax) that are currently allowed. For such an EDFA, the tilt is a function of the gain (e.g. represented by the dgt - dynamic gain tilt) and it is not possible to set a tilt target independently of the gain.

[...]

PR #249, opened 2019-05-27
Do not mix THz and Hz

### LINKED ISSUE #243: Bug in interpolation in Edfa class
The function `interpol_params` calculates DGT, NF ripple and gain ripple at channel frequencies by interpolating values given at other frequencies in the amplifier configuration file. These other frequencies are calculated as follows:

`amplifier_freq = itufl(len(self.params.dgt), self.params.f_min, self.params.f_max) * 1e12 # Hz`

But f_min and f_max are already given in Hz so the multiplication by  `1e12` is wrong. This results in DGT, NF ripple and gain ripple getting constant values corresponding to the lowest frequency represented in the configuration file.

The bug was introduced with PR #222.

jktjkt:
While the `itufl()` function uses THz by default, its values depend on the frequency range that is passed via arguments. In this context, the `f_min` and `f_max` come from the default `SpectralInformation` which is in Hz.

Thanks to @ojnas for reporting this.

fixes #243

### DIFF
--- a/gnpy/core/elements.py
+++ b/gnpy/core/elements.py
@@ -573,7 +573,7 @@ class Edfa(Node):
                 self.channel_freq, self.nf, self.interpol_dgt and self.interpol_gain_ripple
         """
         # TODO|jla: read amplifier actual frequencies from additional params in json
-        amplifier_freq = itufl(len(self.params.dgt), self.params.f_min, self.params.f_max) * 1e12 # Hz
+        amplifier_freq = itufl(len(self.params.dgt), self.params.f_min, self.params.f_max) # Hz
         self.channel_freq = frequencies
         self.interpol_dgt = interp(self.channel_freq, amplifier_freq, self.params.dgt)
 


jktjkt:
@EstherLerouzic , @ojnas : I tried to come up with a test case, but it seems that the existing simple tests such as [`test_fixed_gain_nf`](https://github.com/Telecominfraproject/oopt-gnpy/blob/develop/tests/test_amplifier.py#L88) do not excercise the `Edfa.gprofile`.

Hence I am not sure about how to best test this fix. Do you think that it makes sense to construct a single fixed-gain EDFA and test how different wavelengths are propagated? Should that be put into `tests/test_propagation.py`, or is that file "too high level" for that?

ojnas:
I think it makes more sense to put it in `test_amplifier.py`. Currently there are only tests of NF at a single frequency, i.e. no test of NF ripple, gain ripple or DGT.

jktjkt:
@ojnas , I'm afraid I'll need more help here.

### What exactly is DGT?

I dug into the code and some papers. 

1. According to Hui's and O'Sullivan's [Fiber Optic Measurement Techniques](https://www.elsevier.com/books/fiber-optic-measurement-techniques/hui/978-0-12-373865-3) (whose preview is fortunately on Google Books), DGT refers to the Dynamic Gain Tilt which is a *"[...] slope of small signal gain around a central wavelength [...] in presence of a strong holding light [...]"*. I.e., the effective tilt for small signals when the amplification power is "consumed by a strong signal" if I can paraphrase.

2. GNPy used to carry this definition (58ac717f):
>  DGT_96:  This is the so-called Dynamic Gain Tilt of the EDFA in dB/dB. It is the change in gain at each wavelength corresponding to a 1dB change at the longest wavelength supported.  The value can be obtained experimentally or through analysis of the cross sections or Giles parameters of the Er fibre.  This is experimentally measured by changing the gain of the amplifier above the maximum flat gain while not changing the internal VOA (i.e. the mid-stage VOA is set to minimum and does not change during the measurement). Note that the measurement can change the  gain by an arbitrary amount and divide by the gain change (in dB) which is measured at the reference wavelength (the red end of the band).

3. There's also [your definition](https://github.com/Telecominfraproject/oopt-gnpy/issues/118#issuecomment-433953528) which also makes sense to a layman such as myself, but at this point I just give up :).

### How DGT affects the results

In `elements.py`, the `interpol_dgt` is used in `_gain_profile()`, but it only takes an effect if `tilt_target` is non-zero. All of the existing code, examples, etc use `tilt_target = 0`.

This interpolation is also used for `gain_ripple` and `nf_ripple`. All examples and tests set these values (as in, vectors of performance or calibration data) to zero for some reason.

Now, if I explicitly set `edfa.tilt_target = 0.5`, for example, then the resulting gain profile is indeed tilted (and it is not tilted when the fix for this bug is not present). But I have no idea what I am doing right now, this is too deep into EDFA simulation for now.

Some work-in-progress code which doesn't really work because the actual tilt is different than the set point:
```py
@pytest.mark.parametrize("gain, tilt_target", [(25, 1), (25, 0.3), (22, 1.2)])
def test_fixed_gain_dgt(gain, tilt_target, setup_edfa_fixed_gain, si):
    """=> interpolating the dgt"""
    edfa = setup_edfa_fixed_gain
    frequencies = array([c.frequency for c in si.carriers])
    pin = array([c.power.signal+c.power.nli+c.power.ase for c in si.carriers])
    pin = pin/db2lin(gain)
    baud_rates = array([c.baud_rate for c in si.carriers])
    edfa.operational.gain_target = gain
    edfa.tilt_target = tilt_target
    pref = Pref(0, -gain, lin2db(len(frequencies)))
    edfa.interpol_params(frequencies, pin, baud_rates, pref)
    print(edfa.gprofile)

    assert pytest.approx(edfa.gprofile[-1] - edfa.gprofile[0], abs=0.01) == tilt_target
```

ojnas:
Yes, DGT = Dynamic Gain Tilt. I think the currently implemented EDFA model was introduced in this presentation, but this was before I got involved:

https://telecominfraproject.workplace.com/groups/OOPT.PSE/permalink/957144244450445/

> In elements.py, the interpol_dgt is used in _gain_profile(), but it only takes an effect if tilt_target is non-zero.

Right, because the current implementation assumes a variable gain EDFA with an internal VOA, where the tilt target can be set independently of the gain target.

ggrammel:
Indeed, the link provided by Jonas was the base for DGT definition. From past discussions I recall that the precise way DGT is supposed to be calculated was an issue. As we have the Lumentum data at our disposition, I think it may make sense to describe how to extract from that raw data to a DGT that can be used in GNPy: https://telecominfraproject.workplace.com/groups/OOPT.PSE/permalink/1171444513020416/

ojnas:
This fix really should be merged already.

jktjkt:
fair enough, let's not block this for test coverage

PR #251, opened 2019-05-30
Use exceptions instead of `sys.exit()` when reporting configuration errors

jktjkt:
The code was originally written with direct `exit()` invocation in order to present nice error messages to the user who is running examples from our `examples/` directory. However, this means that one cannot really use the library *as a library*, handling errors in a completely different manner. As an example, if someone wants to write a PCE server with GNPy for impairment calculation, it is most certainly a wrong thing to do to hard-exit an entire daemon when handling a malformed request.

This patch series therefore introduces some rudimentary exception hierarchy, and then we add some pretty-printers for colorful printing in our examples. It already distinguishes between "equipment library errors" (malformed data) and "network topology errors" (nonsensical topologies).

(My editor apparently decided to strip extra whitespace from some offending lines, which is a pretty standard configuration. Sorry for mixing these up, though.)

Cc: @EstherLerouzic

### DIFF
--- a/gnpy/core/elements.py
+++ b/gnpy/core/elements.py
@@ -624,14 +624,8 @@ class Edfa(Node):
             nf_avg = pin_ch - polyval(nf_model.nf_coef, pin_ch) + 58
         elif type_def == 'advanced_model':
             nf_avg = polyval(nf_fit_coeff, -dg)
-        else :
-            print(
-                f'\x1b[1;31;40m'\
-                + f'CRITICAL: unrecognized type def _{self.params.type_def}_\n\
-                    => please check eqpt_config.json'\
-                + '\x1b[0m'
-                )                        
-            exit()            
+        else:
+            assert False, "Unrecognized amplifier type, this should have been checked by the JSON loader"
         return nf_avg+pad, pad
 
     def _calc_nf(self, avg = False):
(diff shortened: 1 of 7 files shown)


gnpy/core/elements.py
```diff
@@ -624,14 +624,8 @@ def _nf(self, type_def, nf_model, nf_fit_coeff, gain_min, gain_flatmax, gain_tar
...
-        else :
-            print(
-                f'\x1b[1;31;40m'\
-                + f'CRITICAL: unrecognized type def _{self.params.type_def}_\n\
-                    => please check eqpt_config.json'\
-                + '\x1b[0m'
-                )                        
-            exit()            
+        else:
+            assert False, "Unrecognized amplifier type, this should have been checked by the JSON loader"
```
EstherLerouzic: It might be usefull to show which  type is invalid here , I think it is self.params.type_def

jktjkt: Thanks for checking this patch series, Esther.

I've used an assertion instead of a regular exception here because this code path will not be ever taken unless our code is buggy. This is "risky" because assertions are turned into a no-op when optimizations are active, so I would prefer not to add more information here -- they might not even be shown.

PR #252, opened 2019-05-31
Bug fixes create eqpt sheet

EstherLerouzic:
Corrections on the create_eqpt_sheet.py program that had a wrong generation when listed links used the node_A name several times for ILAs.
I also simplified the program a lot. 

Note that create_eqpt_sheet.py  is not part of the library, only a easy-to-use tool to help user with the excel file generation.

I also added some coordinates on examples to be able to use the -pl feature


examples/create_eqpt_sheet.py
```diff
@@ -11,64 +11,72 @@
...
+PARSER = ArgumentParser()
+PARSER.add_argument('workbook', nargs='?', default='meshTopologyExampleV2.xls',
+                    help='create the mandatory columns in Eqpt sheet')
+ALL_ROWS = lambda sh, start=0: (sh.row(x) for x in range(start, sh.nrows))
 
-Shortlink = namedtuple('Link', 'src dest')
+class Node:
+    """ Node element contains uid, list of connected nodes and eqpt type
+    """
+    def __init__(self, uid, to_node):
```
jktjkt: This looks like a good candidate for `@dataclass`, but given that this is not the "core library thing", let's ignore this for now.

PR #253, opened 2019-05-31
Capacity planning part1 yang

EstherLerouzic:
adding yang files with detailed explanation in commits
this draft PR to check codacy and codecov reports and do requested changes


EstherLerouzic:
@jktjkt could you please have a look and tell me what to improve ?
unfortunately, github lists the commits ordered by date. the correct order is this one:
(last first, so you should read 11 first :))

1. - Adding a no-path case for test coverage
2. - Adding a test on json response generation
3. - adding a test for the csv creation
4. - minor fixes for spacing and indent
5. - changes to improve quality
6. - Remove @ in path-metric names and extend type to decimal64
7. - update test and corrections according to updated yang
8. - correction of the json generation to map to the path-computation yang model
9. - Yang of the request and response json file to be used by an API
10. - First model for the eqpt-config json
11. - Yang model for the GNPy topology

I think that we could add some automatique verification for yang models as well using pyang (as detailed in commits messages). What do you think ?

EstherLerouzic:
I rebased the branch on the latest develop

EstherLerouzic:
I close this one and one a new one w/o yang models

Issue #100, opened 2018-08-06
Elements class structure


jeanluc-auge:
Can we think of a better __init__ structure than the extensive use of namedtuple we are having?
[init class EDFA](https://github.com/Telecominfraproject/oopt-gnpy/blob/9839681bc05c17c4320f3602f8d8868bbd14afd2/gnpy/core/elements.py#L322)
I have several issues with these namedtuple. For example in the EDFA class:
* the params parameter cannot be skipped because it will require the EDFAparams positional arguments: the None default value in __init__ is not working by the way.
* the params cannot be updated after the calss creation because namedtuple are unmutable. As a result I have (for example): edfa.con_in != edfa.params.con_in in some instances
* most importantly we lost the flexiblity of the initial class structure where there was no need to update the code when adding a new parameter in the json input files. now you need to update several namedtupple and you lose compatibility with previous json files.

dutc:
Discussed on July 31 Coder's Call (on Workplace.) 

No immediate action items.

ggrammel:
2018-10-23: we leave issue closed but expect it to come up soon with a pull request by JL/Esther

PR #257, opened 2019-06-06
Python: do not use a mutable default value

jktjkt:
...and increase the CI coverage as well while I'm at this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,7 @@ install:
 script:
   - pytest --cov-report=xml --cov=gnpy
   - rstcheck --ignore-roles cite --ignore-directives automodule --recursive --ignore-messages '(Duplicate explicit target name.*)' .
+  - cd examples; ./transmission_main_example.py
+  - cd examples; ./path_requests_run.py
 after_success:
   - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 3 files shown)


ojnas:
The change in `class Edfa` in `elements.py` reverts a previous change in commit ff6d81b, which seems related to issue #100 by @jeanluc-auge. Could you maybe comment on what, if anything, has changed?

jktjkt:
> The change in `class Edfa` in `elements.py` reverts a previous change in commit [ff6d81b](https://github.com/Telecominfraproject/oopt-gnpy/commit/ff6d81b749a048ecd99e450064ed376ad121e629), which seems related to issue #100 by @jeanluc-auge. Could you maybe comment on what, if anything, has changed?

It seems that I forgot to comment here, sorry. Adding a reply mainly for future reference.

Thanks for checking this. I looked into the output of `git blame` when I was making this change. I am not sure what Jean-Luc was referring to when writing that the default values do not work. Either way, this is not really a revert of his change, it's rather fixing a Python bug which was accidentally introduced in that commit. The explanation is [in the commit message](https://github.com/Telecominfraproject/oopt-gnpy/pull/257/commits/d8feccc7157389e8f05a79393b292ec23e3819fc).

PR #263, opened 2019-06-18
Raman feature

aleFerrari:
This pull request contains a new object called RamanFiber.
RamanFiber is a copy of the Fiber object which uses a Raman solver to properly estimate the Raman scattering and an NLI solver which takes into account the Raman scattering in the propagation.


ojnas:
In the `__str__` function of RamanFiber, the parameter `self.pch_out_db` is calculated taking only linear loss into account, i.e. not including Raman gain:

`f'  pch out (dBm): {self.pch_out_db!r}'`

So the output when running transmission_with_raman_main_example.py is a bit confusing since it looks like there was no gain in the RamanFiber.

aleFerrari:
> In the `__str__` function of RamanFiber, the parameter `self.pch_out_db` is calculated taking only linear loss into account, i.e. not including Raman gain:
> 
> `f' pch out (dBm): {self.pch_out_db!r}'`
> 
> So the output when running transmission_with_raman_main_example.py is a bit confusing since it looks like there was no gain in the RamanFiber.

You are right.

In the feature Raman feature only the `self.propagate(self, *carriers)` of RamanFiber has been modified, thus only the carriers are updated correctly.
`self.pch_out_db` is instead computed according to the `self.update_pref(self, pref)` method which make use of the `self.loss` property only.

In presence of Raman amplification, the gain/loss is different on each channel, while `self.pch_out_db` is just a number. Thus, it can be fixed extending `self.update_pref(self, pref)` to use a vector `self.pch_out_db`.
@jktjkt, @EstherLerouzic, @ggrammel Does it make conflict with something else?

ojnas:
In power_mode, `pref` is used to calculate effective gain of EDFAs. Currently the gain (and thereby output power) of an EDFA  following a RamanFiber will be too high since calculation of `pref` in RamanFiber does not take Raman gain into account. But I don't think it will work to just replace `self.pch_out_db` with a vector without other changes as well.

ggrammel:
Not that I am aware, I only recall that J.-L. mentioned potential issues with autoconfiguration, but have no more details.

Gert

aleFerrari:
> In power_mode, `pref` is used to calculate effective gain of EDFAs. Currently the gain (and thereby output power) of an EDFA following a RamanFiber will be too high since calculation of `pref` in RamanFiber does not take Raman gain into account. But I don't think it will work to just replace `self.pch_out_db` with a vector without other changes as well.

Ok, so `self.pch_out_db` has to be scalar.
If this number is used to compute the EDFA gain when it is in power_mode, I think the best solution would be to set `self.pch_out_db` equal to the average channel power at the output of the RamanFiber.
@ojnas @jktjkt, @EstherLerouzic, @ggrammel, please let me know if the assumptions are correct and if it could create some conflict with the rest of the code.

jktjkt:
I think that setting `RamanFiber.pch_out_db` to the average value makes sense (and we should probably also rename it so that it says "dbm" and "average"). The actual power of carriers should be correctly updated in the `SpectralInfo` as far as I can tell.

EstherLerouzic:
I also think that the average value makes sense, maybe update ' self.pch_out_db = round(pref.pi - self.loss, 2)',  using the spectral_info.pref already updated with mean carriers.power.signal?

examples/sim_params.json
```diff
@@ -0,0 +1,16 @@
+{
```
jktjkt: I agree that there's merit in splitting out "simulation parameters" from "topology of the network", but I do not think that a pull request implementing the Raman feature is a good place to change this. There's the `SpectralInformation` stanza in the `example_network.json` which already contains these "global" information. I would prefer that structure to be used instead of a new `sim_params.json`.

examples/sim_params.json
```diff
@@ -0,0 +1,16 @@
...
+  "list_of_channels_under_test": [1, 18, 37, 56, 75],
+  "raman_parameters": {
+    "flag_raman": true,
```
jktjkt: I checked minutes from 2019-05-28 before I wrote this comment, so I know that there's a difference between "use Raman engine" and "turn on Raman amplification", but I cannot map the result of that discussion into this JSON file.

My understanding is that it's possible to "use the Raman engine" with no Raman pumps being active (right?). How do I do that in terms of input JSON files?

The way I understand it, Raman pumps are controlled by using a `RamanFiber` in the network topology **and** listing the pumps in that file as well.

- What happens when I put a `RamanFiber` with no pumps into the JSON network file?
- What happens when I use `raman_paramaters` in the sim input, but with `flag_raman` set to `False`?

examples/sim_params.json
```diff
@@ -0,0 +1,16 @@
...
+    "space_resolution": 10e3,
+    "tolerance": 1e-8,
+    "verbose": 2
```
jktjkt: If this is about a debugging verbosity, let's control this via an option to the program (and an appropriate log level).

examples/transmission_with_raman_main_example.py
```diff
@@ -0,0 +1,304 @@
...
+parser.add_argument('-pl', '--plot', action='store_true')
+parser.add_argument('-v', '--verbose', action='count', default=0, help='increases verbosity for each occurence')
+parser.add_argument('-l', '--list-nodes', action='store_true', help='list all transceiver nodes')
+parser.add_argument('-po', '--power', default=0, help='channel ref power in dBm')
+parser.add_argument('-names', '--names-matching', action='store_true', help='display network names that are closed matches')
+parser.add_argument('filename', nargs='?', type=Path,
+                    default=Path(__file__).parent / 'raman_edfa_example_network.json')
+parser.add_argument('source', nargs='?', help='source node')
+parser.add_argument('destination',   nargs='?', help='destination node')
+
```
jktjkt: When my other comments are implemented, this file ends up very similar to the existing `transmission_main_example.py`. My recommendation is therefore to control selection of the Raman vs. "traditional" engine just by the choice of the input JSON file ("does it contain `RamanFiber`?").

Looking at how the code looks right now, I don't see any possible regression. We don't have that concept of "Raman amplifiers" yet, so autodesign is not affected as far as I could tell.

gnpy/core/elements.py
```diff
@@ -417,6 +417,201 @@ def __call__(self, spectral_info):
...
+        return pref._replace(p_span0=pref.p0, p_spani=self.pch_out_db)
+
+    def __call__(self, spectral_info):
+        self.carriers_in = spectral_info.carriers
+        carriers = tuple(self.propagate(*spectral_info.carriers))
+        pref = self.update_pref(spectral_info.pref, *carriers)
+        self.carriers_out = carriers
+        return spectral_info.update(carriers=carriers, pref=pref)
+
+    def propagate(self, *carriers):
```
jktjkt: Apart from a small difference in the constructor, this `propagate` method (and one difference above) is the only difference between classes `Fiber` and `RamanFiber` (as long as I'm reading the diff correctly, accounting for some cleanup in the develop branch). I would prefer to have `RamanFiber` derived from `Fiber`, overriding the `propagate` method and the constructor. This will reduce code duplication; in general, copying code around is something that should be avoided.

gnpy/core/elements.py
```diff
@@ -417,6 +417,201 @@ def __call__(self, spectral_info):
...
+                     * (carrier.power.signal/carrier.baud_rate) * psi
+
+        g_nli *= (16 / 27) * (self.gamma * self.effective_length)**2 \
+                 / (2 * pi * abs(self.beta2()) * self.asymptotic_length)
+
+        carrier_nli = carrier.baud_rate * g_nli
+        return carrier_nli
+
+    def update_pref(self, pref, *carriers):
+        pch_out_db = 10*log10(mean([carrier.power.signal for carrier in carriers])) + 30
```
jktjkt: This got introduced in 1a1346461be28f37538c5dbbf500b797add3b30e, but I do not fully understand it.

I remember that we agreed on using a mean of all carried channels, but I don't understand the `+ 30` bit.

gnpy/core/science_utils.py
```diff
@@ -0,0 +1,842 @@
...
+        return carrier_nli
+
+    def _compute_eta_matrix(self, carrier_cut, *carriers):
+        cut_index = carrier_cut.channel_number - 1
+        # Matrix initialization
+        matrix_size = max(carriers, key=lambda x: getattr(x, 'channel_number')).channel_number
+        eta_matrix = np.zeros(shape=(matrix_size, matrix_size))
+
+        # SPM
+        if self.nli_params.verbose:
```
jktjkt: This (and everywhere else) should be converted to use the `logging` module. That way, the code just emits logging messages with a given severity (error/warn/info/debug/trace) and "component" (JSON parsing, Raman solver, PCE, etc), and the framework selects what to capture and what messages to discard.

gnpy/core/science_utils.py
```diff
@@ -0,0 +1,842 @@
...
+        for interfering_carrier in carriers:
+            g_interfearing = interfering_carrier.power.signal / interfering_carrier.baud_rate
+            g_signal = carrier.power.signal / carrier.baud_rate
+            g_nli += g_interfearing**2 * g_signal * self._psi(carrier, interfering_carrier)
+        g_nli *= (16.0 / 27.0) * (gamma * effective_length)**2 /\
+                 (2 * np.pi * abs(beta2) * asymptotic_length)
+        carrier_nli = carrier.baud_rate * g_nli
+        return carrier_nli
+
+    def _psi(self, carrier, interfering_carrier):
```
jktjkt: This is already in `gnpy/core/elements.py`. I'm all for moving all "real computation" to a dedicated submodule, but I would prefer to avoid a state where stuff got duplicated.

My preference would be to move these methods one-by-one (or at least separately from the NLI solver as done in commit f9bd6310f182b62f9d373f4cbbba11b0af312b9d).

gnpy/core/science_utils.py
```diff
@@ -0,0 +1,842 @@
...
+        integrand_f1 = np.zeros(len(f1_array))
+        for f1_index, f1 in enumerate(f1_array):
+            delta_beta = 4 * np.pi**2 * (f1 - f_eval) * (f2_array - f_eval) * \
+                         (beta2 + np.pi * beta3 * (f1 + f2_array - 2 * f_ref_beta))
+            integrand_f2 = self._generalized_rho_nli(delta_beta, rho_norm_pump, z, alpha0)
+            integrand_f1[f1_index] = 2 * np.trapz(integrand_f2, f2_array)  # 2x since integrand_f2 is symmetric in f2
+        generalized_psi = 0.5 * sum(integrand_f1) * pump_carrier.baud_rate
+        return generalized_psi
+
+    def _generalized_psi(self, carrier_cut, pump_carrier, f_eval, f_cut_resolution, f_pump_resolution):
```
jktjkt: This function and the one above are very similar (they even have the exact same docstring!). Please describe them in more detail; the docstring should explain what tradeoffs are in between them, etc. A pointer to a literature is perfect, and a layman's explanation ("for Jan") would be much appreciated. E.g., "The fast method only computes the slope of Foo Bar using the two extreme points instead of going via all carrier frequencies", or whatever :).

The common parts are perhaps just the syntax noise. For example, half of the code just copies variables such as `self.fiber_params.beta2` into `beta2`, a variable which only gets used once. I'm not sure what to do here (if anything) -- it's probably true that using just `beta2` in an expression looks neater, but I'm not sure if it's worth the boilerplate above (which gets copied twice into the fast method).

examples/transmission_with_raman_main_example.py
```diff
@@ -0,0 +1,304 @@
...
+    print('\n'.join([f'Power mode is set to {power_mode}',
+                     f'=> it can be modified in eqpt_with_raman_config.json - Span']))
+
+    pref_ch_db = lin2db(req.power*1e3) #reference channel power / span (SL=20dB)
+    pref_total_db = pref_ch_db + lin2db(req.nb_channel) #reference total power / span (SL=20dB)
+    build_network(network, equipment, pref_ch_db, pref_total_db)
+    configure_network(network, sim_params)
+    path = compute_constrained_path(network, req)
+
+    spans = [s.length for s in path if isinstance(s, RamanFiber)]
```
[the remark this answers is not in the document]
jktjkt: This one was, apparently, just for mentioning the total span length, but I fixed it anyway.

gnpy/core/elements.py
jktjkt: I missed the fact that `carrier.power.signal` is in W, not mW, and therefore the +30 is just a conversion.

examples/sim_params.json
jktjkt: I made a mistake here, I missed the fact that `SpectralInfromation` is currently defined in the equipment library, not in the network topology JSON file. Sorry for my confusion. Let's take this and merge it as an improvement.

(I'll blame this on my recent YANG work where this is different.)

Issue #260, opened 2019-06-14
Gnpy docker


jeanluc-auge:
There are several obstacles for newcomers to the project if they are not developpers:
* clone or download the project from github
* have a recent python version 3.6 and soon 3.7: this is sometimes difficult with older os distributions
* have the right environment: requirement.txt and correct python path

The proposed solution is to provide a Dockerfile and a docker-compose.yml to build a gnpy image with the right environnement and run gnpy inside the container. A shared volume (folder) between the host and the container enables the file exchange (topology and configuration json/xls files)

This is typically proposed on other projects, see for example [ATC](https://github.com/apache/trafficcontrol/tree/master/infrastructure/docker) which my new job get me working on ;) as a standalone and easier setup.

I will push next week the dockerfiles needed to build and run a gnpy image with some documentation. However there are 2 possibilities:
*option 1: The dockerfile configuration downloads the source code from GitHub (branch develop for example) directly to the container: this is very easy for newcomers, they don’t have anyhting to deal with Github, except copy past the dockerfiles.
*option 2: The dockerfile configuration copies the source code from the local drive into the container: this means that the user already cloned the Git repo. But it gives more control for developers because they can modify the code locally before building the container to test the code. This 2nd option is used in ATC.

Could you comment and tell me what you prefer? 
And do you think it can provide value ?

jktjkt:
I think that it makes a lot of sense to provide a prepackaged tool for newcomers to use. Docker looks like a great match for this because it has a big momentum of users, and the majority of users like it that way.

The situation is a bit different with developers. I think that Python has great built-in tools for managing virtual environments, and setting one up is a matter of running one simple command. I do not see much added value with Docker here (well, unless people are stuck on something like RHEL6 for their work machine). Someone who works with GNPy should understand Python reasonably, and I think that `venv` is a part of that understanding. If a fellow developer wants to run Docker to make their life easier, be my guest of course -- but to me, it does not look like a great use of resources to create a container first, and then set up a venv inside. That's one layer too many for my personal taste.

This means that, from my point of view, this is how a perfect solution should look like:

- It's a one-click thing for users. No need to set up anything.
- We have it under our control. This means CI coverage and no manual steps.

I think you know that, but Esther said that someone from Orange (but not you) will come talk about how they already packaged PSE with Docker. Just making sure that there's no duplicate effort here, but you are probably aware of that activity.

ojnas:
[...]

PR #265, opened 2019-06-19
docker files

jeanluc-auge:
see README.md and issue #260

### DIFF
--- a/docker/Dockerfile-compose
+++ b/docker/Dockerfile-compose
@@ -1,0 +1,25 @@
+
+FROM python:3.6
+
+#COPY . /app
+
+#RUN pip install -r /app/requirements.txt
+
+#ENV PYTHONPATH=/app
+
+#WORKDIR /app/examples
+
+ARG BRANCH="develop"
+
+ADD https://github.com/Telecominfraproject/oopt-gnpy/archive/${BRANCH}.zip /tmp
+
+RUN unzip /tmp/develop.zip -d /
+
+RUN pip install -r /oopt-gnpy-develop/requirements.txt
+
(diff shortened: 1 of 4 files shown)


ojnas:
I tested it on Docker for Windows and it works as described, nice!

Having a shared folder is convenient but it would be even more convenient if the examples folder could be shared so that existing input files, etc., could be edited directly from the host. Alternatively add some description in the README how to copy files from the example folder to the shared folder and work from there.

ojnas:
Here is an alternative approach that shares the example folder. This doesn't use docker-compose but I think the same approach could be used for that case.

1. Download this [Dockerfile](https://github.com/ojnas/oopt-gnpy/blob/docker/docker/Dockerfile).
2. Change to the directory with the downloaded Dockerfile.
3. Build the docker image with:
`docker build -t gnpy .`
4. Run the docker container with one of the following commands depending on platform:
For Windows using PowerShell:
`docker run -it --rm -v ${PWD}:/oopt-gnpy-develop/shared/ gnpy`
For Linux (and I think also Mac):
` docker run -it --rm -v $(pwd):/oopt-gnpy-develop/shared/ gnpy`
5. In the resulting bash prompt, run e.g. `python3 transmission_main_example.py`.

This approach should create an examples folder in the same directory as the Dockerfile. Changes to any file in this folder, e.g. eqpt_config.json, should be seen also in the container and should persist when stopping and restarting the container.

jeanluc-auge:
Thanks Jonas, I pushed a new version

jktjkt:
I'm closing this in favor of #295. Thanks for the code, it helped me get the full automation running!

Issue #227, opened 2019-05-14
amplifier model revision


jeanluc-auge:
Following an email exchange with @ojnas it would be more clear and pratical to separate the spectral model from the other description. Now most of the models have a default_edfa_config.json (all ripple description to 0), so it’s just a minor code change to be able to enter a custom “advanced_config_from_json” input for all models. If it’s left empty, the default config could be assumed for example.
This way we could end up with just 4 models (5 with upcoming Raman):

- *polynomial NF*: can also be used for the fixed gain model with coef [0,0,0, NF], 
so the current fixed_gain model can be deprecated
```json
type_def: “polynomial_NF”, 
gain_flatmax
gain_min
p_max
nf_coef
spectral_info_json (optional)
out_voa_auto
allowed_for_design
```
- *polynomial OSNR* = current unchanged “openroadm” description, you may want to keep type_def: “openroadm” for clarity
```json
type_def: “polynomial_OSNR”
gain_flatmax
gain_min
p_max
osnr_coef
spectral_info_json (optional)
out_voa_auto
allowed_for_design
```

- *operator_model*, which is the current variable gain where you enter NF_min and NF_max instead of a polynomial. => the difference with a polynomial model is minor: you can do the test, compare the NF from the variable gain model with a polynomial curve once you have entered the NF_min and NF_max values of the polynomial range in the variable gain model. I always find < 0.5dB NF at worse (which is <0.2dB SNR at the optimum power) and most of the time <0.1dB NF difference.

```json
type_def: “operator_model”
gain_flatmax
gain_min
p_max
nf_min
nf_max
spectral_info_json (optional)
out_voa_auto
allowed_for_design
```

- *dual_stage*, current unchanged description, this is very handy to support dual stage configurations. It works very well for auto-design, and I have validated it with multiple import from supplier designs: it does the job. It describes a combination of any of the above models.
```json
type_def:”dual_stage”
gain_min (auto-design interest, don’t use below this gain)
raman (auto_design interest: true/false raman in the preamp)
preamp_variety
booster_variety
allowed_for_design
```
```json
type_def:”raman”?
pumps_power_list [p1,p2,p3…]
pumps_frequency_list [f1,f2,f3…]
…
```

And that’s it. For each of these models (except the dual stage of course) you just need to add an optional json input field to load the detailed spectrum information if needed.

jktjkt:
I like this, and it seems that @ojnas likes this, too. Is someone against this proposal?

[...]

PR #266, opened 2019-06-19
YANG models for inventory management and topology description

jktjkt:
Right now, the documentation and code disagree on which values are mandatory, what exactly they mean, etc. Let's try to use a set of YANG models for verification.

## FIXME: What is missing

- [ ] ROADMs
- [ ] EDFA parameters
  - [ ] TODO: implement DGT (et al) indirection
  - [x] check that this set of EDFA parameters makes sense for each possible simulation parameter set
  - [x] rework for #227
  - [x] EDFA maximal gain -- this one doesn't appear to be used by the code anywhere!, which confuses me
  - [x] Raman support (preliminary, on-par with what we already have now in `devel`)
- [x] Fibers
  - [ ] Fiber.dispersion and Fiber.gamma range validation. Also, how do we describe Fiber type's gamma?
  - [x] Fiber operational data
- [x] Transponders
  - [x] Transceiver's tx roll-off; how exactly is this defined? I only understand it "intuitively".
  - [ ] minimal required OSNR (#298)
- [x] topology
  - [x] use [`ietf-network-topology`](https://tools.ietf.org/html/rfc8345#section-4.2) for topology (this is *not* about the [impariment-aware CCAMP work](https://github.com/younglee-ietf/ietf-optical-impairment-yang))
  - [ ] More XPath bits for topology validation -- connecting just those elements which make sense together
- [ ] replace `SpectralInfo`
  - [ ] default transponder for `transmission_main_example`
  - [ ] unclear how to reconcile the `f_min`/`f_max` of the SI and TX
  - [ ] default spectrum allocation (autodesign assumes full allocation -> cannot use transponders)
- [ ] CI coverage:
  - [ ] YANG validation & linting
  - [ ] positive and negative JSON samples for constraint testing
    - "ignoring the YANG-specificity":
      - use [this `jsonschema` plugin to `pyang`](https://github.com/cmoberg/pyang-json-schema-plugin/blob/master/jsonschema/jsonschema.py) or [this `jsonschema` plugin for `pyang`](https://github.com/OpenNetworkingFoundation/EagleUmlCommon/blob/ToolChain/YangJsonTools/json_schema.py)
      - feed their results into [jsonschema](https://python-jsonschema.readthedocs.io/en/stable/) for actual JSON validation
    - **or** just use `yanglint` proper, because it works (but might be a bit painful to package for CI and users),
    - [ ] **or** use [yangson](https://github.com/CZ-NIC/yangson), a native Python library
      - [ ] prepare the JSON data in the `ietf-yang-library` format
- [ ] input validation
  - this one should be in Python, not via `yanglint` for dependency reasons
- [ ] conversion scripts
- [ ] use this as an input format

## Equipment Library

The first step in adding YANG description for gnpy's input is the equipment library (`tip-photonic-equipment`). It contains data about all defined EDFA and Fiber types. This is supposed to be functionally equivalent to the `equpt_config.json`. The JSON structure has changed because I (much) value YANG readability over being directly compatible with our previous format (there will be a conversion script anyway).

The core idea of this model is to describe capabilities of the simulation engine as it exists, which means that the individual choice/case statements mirror our different "simulation input parameters". The user is not expected to do any augmentations -- just describe the amplifiers, fiber, etc, with data.

The pre-YANG code actually split stuff from `equpt_config.json` into additional JSON files for "fancy bits", such as the DGT LUT. That's something that, IMHO, does not make sense when we're willing to ship with machine-validation of the complete input set. So instead of deferring to another JSON file for the NF-/gain-ripple/DGT, let's move everything in-line into the input data. This has one obvious downside in making the amplifier data a bit too verbose. There were several options:

- Ignore the human-friendliness and push everything into the amplifier description. This is nice and self-contained, but the data are going to be very, very long, and the majority of the WG was worried that it would make human editing too difficult.

- Move everything to a side-loaded JSON file. This option separates out some numerical parameters from the equipment library, and therefore splits the configuration into two places. One of these places would be exempt from the YANG validation, and loaded via unspecified means.  That's a no-go.

- Put stuff into a YANG model, but use one level of indirection between the amplifier description and the numerical data.

As a compromise, we've chosen the last option.

In the real world, some "common fiber types" are well-defined by ITU, such as the SSMF. Esther tried to model this via a set of identities and YANG `identityref`s. I think that there's no disadvantage in shipping these data as a default content of the YANG-formatted datastore, similarly to how the equipment library used to be structured prior to this patch. Once again, I'm following the pattern where the user can change any *data* without augmenting the YANG model. The only reason for editing/augmenting the (equipment) YANG model should be changes in our simulation *engine*, such as when adding different input parameters for NF calculations, or adding Raman amplification, etc.

The amplifier model has been reworked a bit (only in the YANG model for now, though). I've reduced the number of available "simulation parameters" to a reasonable minimum as suggested by Jean-Luc (cf. issue #227):

- a polynomial NF model
- a polynomial OSNR model ("OpenROADM")
- a simplified model for operators with NF_min and NF_max
- a dual-stage amplifier using any of the above

## Topology Model

The topology model (`tip-photonic-topology`) uses `leafref`s for "instantiating" actual nodes from models defined in the equipment library. The topology is unidirectional.

At first, I used an ad-hoc, custom topology for simplicity. This was changed in response to Jonas' comment; there's clearly no need to reinvent this particular wheel. Now the model builds on top of RFC8345.
    
Both [`ietf-network-topology`](https://tools.ietf.org/html/rfc8345#section-4.2) and the associated `ietf-network` are needed.  The augmentations make it a bit harder to see the YANG rendering of the resulting modules, but that's just how the IETF model was designed.  There's nothing to fix on our side.

Unlike the current JSON, Fibers are not nodes anymore. They are implemented as `nt:link` augmentations.

## Sample Data

To see how the (total, complete) input data to GNPy might look like, I'm including [a simple JSON file]( https://github.com/Telecominfraproject/oopt-gnpy/pull/266/files?file-filters%5B%5D=.json) which defines some amplifiers, fiber, transponders, etc., and uses them to build a very simple path. I think that there's plenty of stuff which is missing, but it shows something which passes the YANG validation.

### Equipment Inventory
```
module: tip-photonic-equipment
  +--rw amplifier* [type]
  |  +--rw type              string
  |  +--rw (simulation)
  |  |  +--:(polynomial-NF)
  |  |  |  +--rw polynomial-NF
  |  |  |     +--rw gain-min                      gain
  |  |  |     +--rw gain-flatmax                  power
  |  |  |     +--rw gain-max-extended             power
  |  |  |     +--rw max-power-out                 power
  |  |  |     +--rw gain-ripple*                  db-ratio
  |  |  |     +--rw nf-ripple*                    db-ratio
  |  |  |     +--rw dynamic-gain-tilt*            db-ratio
  |  |  |     +--rw nf-polynomial-coefficients
  |  |  |        +--rw a    polynomial-coefficient
  |  |  |        +--rw b    polynomial-coefficient
  |  |  |        +--rw c    polynomial-coefficient
  |  |  |        +--rw d    polynomial-coefficient
  |  |  +--:(polynomial-OSNR)
  |  |  |  +--rw polynomial-OSNR
  |  |  |     +--rw gain-min                        gain
  |  |  |     +--rw gain-flatmax                    power
  |  |  |     +--rw gain-max-extended               power
  |  |  |     +--rw max-power-out                   power
  |  |  |     +--rw gain-ripple*                    db-ratio
  |  |  |     +--rw nf-ripple*                      db-ratio
  |  |  |     +--rw dynamic-gain-tilt*              db-ratio
  |  |  |     +--rw osnr-polynomial-coefficients
  |  |  |        +--rw a    polynomial-coefficient
  |  |  |        +--rw b    polynomial-coefficient
  |  |  |        +--rw c    polynomial-coefficient
  |  |  |        +--rw d    polynomial-coefficient
  |  |  +--:(min-max-NF)
  |  |  |  +--rw min-max-NF
  |  |  |     +--rw gain-min             gain
  |  |  |     +--rw gain-flatmax         power
  |  |  |     +--rw gain-max-extended    power
  |  |  |     +--rw max-power-out        power
  |  |  |     +--rw gain-ripple*         db-ratio
  |  |  |     +--rw nf-ripple*           db-ratio
  |  |  |     +--rw dynamic-gain-tilt*   db-ratio
  |  |  |     +--rw nf-min               noise-figure
  |  |  |     +--rw nf-max               noise-figure
  |  |  +--:(raman)
  |  |  |  +--rw raman
  |  |  |     +--rw gain    gain
  |  |  |     +--rw nf      noise-figure
  |  |  +--:(dual-stage)
  |  |     +--rw dual-stage!
  |  |        +--rw preamp      -> /tip-photonic-equipment:amplifier/type
  |  |        +--rw booster     -> /tip-photonic-equipment:amplifier/type
  |  |        +--rw gain-min    power
  |  +--rw frequency-min?    frequency <191.35>
  |  +--rw frequency-max?    frequency <196.1>
  |  +--rw has-output-voa?   boolean <false>
  +--rw fiber* [type]
  |  +--rw type          string
  |  +--rw dispersion    decimal64
  |  +--rw gamma         decimal64
  +--rw transceiver* [type]
     +--rw type             string
     +--rw frequency-min?   frequency <191.35>
     +--rw frequency-max?   frequency <196.1>
     +--rw mode* [name]
        +--rw name             string
        +--rw baud-rate        uint64
        +--rw required-osnr    db-ratio
        +--rw tx-osnr          db-ratio
        +--rw grid-spacing     frequency
        +--rw tx-roll-off?     decimal64
        +--rw cost?            uint32 <1>
```
### Topology
```
module: tip-photonic-topology

  augment /ietf-network:networks/ietf-network:network/ietf-network:network-types:
    +--rw photonic-topology!
  augment /ietf-network:networks/ietf-network:network/ietf-network-topology:link:
    +--rw fiber
       +--rw type      -> /tip-photonic-equipment:fiber/tip-photonic-equipment:type
       +--rw length    decimal64
  augment /ietf-network:networks/ietf-network:network/ietf-network:node:
    +--rw (element)
    |  +--:(amplifier)
    |  |  +--rw amplifier
    |  |     +--rw model          -> /tip-photonic-equipment:amplifier/tip-photonic-equipment:type
    |  |     +--rw gain-target    tip-photonic-equipment:gain
    |  |     +--rw tilt-target    tip-photonic-equipment:db-ratio
    |  +--:(concentrated-loss)
    |  |  +--rw concentrated-loss
    |  |     +--rw attenuation?   tip-photonic-equipment:db-ratio <0.0>
    |  +--:(transceiver)
    |     +--rw transceiver
    |        +--rw model    -> /tip-photonic-equipment:transceiver/tip-photonic-equipment:type
    +--rw geo-location
    |  +--rw reference-frame
    |  |  +--rw alternate-system?    string {ietf-geo-location:alternate-systems}?
    |  |  +--rw astronomical-body?   string <earth>
    |  |  +--rw geodetic-system
    |  |     +--rw geodetic-datum?    string <wgs-84>
    |  |     +--rw coord-accuracy?    decimal64
    |  |     +--rw height-accuracy?   decimal64
    |  +--rw (location)?
    |  |  +--:(ellipsoid)
    |  |  |  +--rw latitude     ietf-geo-location:degrees
    |  |  |  +--rw longitude    ietf-geo-location:degrees
    |  |  |  +--rw height?      decimal64
    |  |  +--:(cartesian)
    |  |     +--rw x    decimal64
    |  |     +--rw y    decimal64
    |  |     +--rw z?   decimal64
    |  +--rw velocity
    |  |  +--rw v-north?   decimal64
    |  |  +--rw v-east?    decimal64
    |  |  +--rw v-up?      decimal64
    |  +--rw timestamp?         ietf-yang-types:date-and-time
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ script:
   - ./examples/path_requests_run.py
   - ./examples/transmission_main_example.py examples/raman_edfa_example_network.json --sim examples/sim_params.json --show-channels
   - sphinx-build docs/ x-throwaway-location
+  #- yanglint -t config --strict --merge -f xml -D yang/tip-*.yang yang/all.json
 after_success:
   - bash <(curl -s https://codecov.io/bash)
 jobs:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,9 @@ script:
   - ./examples/transmission_main_example.py examples/raman_edfa_example_network.json --sim examples/sim_params.json --show-channels
   - sphinx-build docs/ x-throwaway-location
   #- yanglint -t config --strict --merge -f xml -D yang/tip-*.yang yang/all.json
+  # ^^ this is with CESNET's libyang. Alternatively, use CZ-NIC's yangson:
+  #- python path/to/yangson/tools/python/mkylib.py yang/ > LIB.json
+  #- yangson -p yang --validate yang/all.json LIB.json
 after_success:
   - bash <(curl -s https://codecov.io/bash)
 jobs:
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ scipy>=1.3.0,<2
 Sphinx>=2.1.1,<3
 sphinxcontrib-bibtex>=0.4.2,<1
 xlrd>=1.2.0,<2
+yangson>=1.3.40,<2
--- a/yang/all.json
+++ b/yang/all.json
@@ -95,5 +95,73 @@
     {
       "type": "Voyager"
     }
-  ]
+  ],
+
+  "ietf-network:networks": {
+    "network": [
+      {
+        "network-id": "TIP OOPT-PSE sample topology",
+        "network-types": {
+          "tip-photonic-topology:photonic-topology": {
+          }
+        },
+        "node": [
+          {
+            "node-id": "trx-1",
+            "tip-photonic-topology:transceiver": {
+              "model": "Voyager"
+            },
+            "tip-photonic-topology:location": {
+              "name": "A"
+            }
+          },
+          {
+            "node-id": "trx-2",
+            "tip-photonic-topology:transceiver": {
+              "model": "Voyager"
+            },
+            "tip-photonic-topology:location": {
+              "name": "B"
+            }
+          },
+          {
+            "node-id": "adfa-A",
+            "tip-photonic-topology:amplifier": {
+              "model": "fixed-22",
+              "gain-target": 19.0,
+              "tilt-target": 10.0
+            }
+          }
+        ],
+        "ietf-network-topology:link": [
+          {
+            "link-id": "fiber trx-1 edfa-1",
+            "source": {
+              "source-node": "trx-1"
+            },
+            "destination": {
+              "dest-node": "edfa-A"
+            },
+            "tip-photonic-topology:fiber": {
+              "type": "SSMF",
+              "length": 60
+            }
+          },
+          {
+            "link-id": "fiber edfa-A trx-2",
+            "source": {
+              "source-node": "edfa-A"
+            },
+            "destination": {
+              "dest-node": "trx-2"
+            },
+            "tip-photonic-topology:fiber": {
+              "type": "SSMF",
+              "length": 50
+            }
+          }
+        ]
+      }
+    ]
+  }
 }
--- a/yang/all.json
+++ b/yang/all.json
@@ -1,0 +1,99 @@
+{
+  "tip-photonic-equipment:amplifier": [
+    {
+      "type": "fixed-27",
+      "polynomial-NF": {
+        "gain-min": 27,
+        "gain-flatmax": 27,
+        "gain-max-extended": 27,
+	"nf-polynomial-coefficients": {
+          "a": 0,
+          "b": 0,
+          "c": 0,
+          "d": 4.9
+	},
+        "max-power-out": 21.9
+      }
+    },
+    {
+      "type": "fixed-22",
+      "polynomial-NF": {
+        "gain-min": 22,
+        "gain-flatmax": 22,
+        "gain-max-extended": 22,
+	"nf-polynomial-coefficients": {
+          "a": 0,
+          "b": 0,
+          "c": 0,
+          "d": 4.8
+	},
+        "max-power-out": 21.9
+      }
+    },
+    {
+      "type": "vg-15-30",
+      "min-max-NF": {
+        "nf-min": 4.9,
+        "nf-max": 12.0,
+        "gain-min": 15,
+        "gain-flatmax": 27,
+        "gain-max-extended": 28,
+        "gain-ripple": [0],
+        "nf-ripple": [0],
+        "dynamic-gain-tilt": [0, 2.4],
+        "max-power-out": 23
+      }
+    },
+    {
+      "type": "vg-10-19",
+      "min-max-NF": {
+        "nf-min": 5.1,
+        "nf-max": 5.8,
+        "gain-min": 10,
+        "gain-flatmax": 17,
+        "gain-max-extended": 20,
+        "gain-ripple": [0],
+        "nf-ripple": [0],
+        "dynamic-gain-tilt": [0, 1.7],
+        "max-power-out": 23
+      }
+    },
+    {
+      "type": "dual--vg-15-30--vg-10-19",
+      "dual-stage": {
+        "preamp": "vg-15-30",
+        "booster": "vg-10-19",
+        "gain-min": 25.0
+      }
+    },
+    {
+      "type": "raman",
+      "raman": {
+        "gain": 12,
+        "nf": 3
+      }
+    }
+  ],
+  "tip-photonic-equipment:fiber": [
+    {
+      "type": "SSMF",
+      "dispersion": 1.67e-05,
+      "gamma": 0.00127
+    },
+    {
+      "type": "NZDF",
+      "dispersion": 0.5e-05,
+      "gamma": 0.00146
+    },
+    {
+      "type": "LOF",
+      "dispersion": 2.2e-05,
+      "gamma": 0.000843
+    }
+  ],
+  "tip-photonic-equipment:transceiver": [
+    {
+      "type": "Voyager"
+    }
+  ]
+}
--- a/yang/all.json
+++ b/yang/all.json
@@ -111,17 +111,15 @@
             "tip-photonic-topology:transceiver": {
               "model": "Voyager"
             },
-            "tip-photonic-topology:location": {
-              "name": "A"
+            "tip-photonic-topology:geo-location": {
+              "latitude": 0,
+              "longitude": 0
             }
           },
           {
             "node-id": "trx-2",
             "tip-photonic-topology:transceiver": {
               "model": "Voyager"
-            },
-            "tip-photonic-topology:location": {
-              "name": "B"
             }
           },
           {
--- a/yang/ietf-geo-location@2019-02-17.yang
+++ b/yang/ietf-geo-location@2019-02-17.yang
@@ -1,0 +1,251 @@
+module ietf-geo-location {
+  namespace "urn:ietf:params:xml:ns:yang:ietf-geo-location";
+  prefix geo;
+  import ietf-yang-types { prefix types; }
+
+  organization
+    "IETF NETMOD Working Group (NETMOD)";
+  contact
+    "Christian Hopps <chopps@chopps.org>";
+
+  // RFC Ed.: replace XXXX with actual RFC number and
+  // remove this note.
+
+  description
+    "This module defines a grouping of a container object for
+     specifying a location on or around an astronomical object (e.g.,
+     The Earth).
+
+     Copyright (c) 2018 IETF Trust and the persons identified as
+     authors of the code. All rights reserved.
+
+     Redistribution and use in source and binary forms, with or
+     without modification, is permitted pursuant to, and subject to
+     the license terms contained in, the Simplified BSD License set
+     forth in Section 4.c of the IETF Trust's Legal Provisions
+     Relating to IETF Documents
+     (https://trustee.ietf.org/license-info).
+
+     The key words 'MUST', 'MUST NOT', 'REQUIRED', 'SHALL', 'SHALL
+     NOT', 'SHOULD', 'SHOULD NOT', 'RECOMMENDED', 'NOT RECOMMENDED',
+     'MAY', and 'OPTIONAL' in this document are to be interpreted as
+     described in BCP 14 [RFC2119] [RFC8174] when, and only when,
+     they appear in all capitals, as shown here.
+
+     This version of this YANG module is part of RFC XXXX
+     (https://tools.ietf.org/html/rfcXXXX); see the RFC itself for
+     full legal notices.";
+
+  // RFC Ed.: replace XXXX with actual RFC number and
+  // remove this note.
+
+  revision 2019-02-17 {
+    description "Initial Revision";
+    reference "RFC XXXX: YANG Geo Location";
+  }
+
+  typedef degrees {
+    type decimal64 {
+      fraction-digits 16;
+    }
+    units "decimal degrees";
+    description "Coordinate value.";
+  }
+
+  feature alternate-systems {
+    description
+      "This feature means the device supports specifying locations
+       using alternate systems for reference frames.";
+  }
+
+  grouping geo-location {
+   description
+      "Grouping to identify a location on an astronomical object.";
+
+    container geo-location {
+      description
+        "A location on an astronomical body (e.g., The Earth)
+         somewhere in a universe.";
+
+      container reference-frame {
+        description
+          "The Frame of Reference for the location values.";
+
+        leaf alternate-system {
+          if-feature alternate-systems;
+          type string;
+          description
+            "The system in which the astronomical body and
+             geodetic-datum is defined. Normally, this value is not
+             present and the system is the natural universe; however,
+             when present this value allows for specifying alternate
+             systems (e.g., virtual realities). An alternate-system
+             modifies the definition (but not the type) of the other
+             values in the reference frame.";
+        }
+        leaf astronomical-body {
+          type string {
+            pattern '[ -@\[-\^_-~]*';
+          }
+          default "earth";
+          description
+            "An astronomical body as named by the International
+             Astronomical Union (IAU) or according to the alternate
+             system if specified. Examples include 'sun' (our star),
+             'earth' (our planet), 'moon' (our moon), 'enceladus' (a
+             moon of Saturn), 'ceres' (an asteroid),
+             '67p/churyumov-gerasimenko (a comet). The value should
+             be comprised of all lower case ASCII characters not
+             including control characters (i.e., values 32..64, and
+             91..126). Any preceding 'the' in the name should not
+             be included.";
+        }
+        container geodetic-system {
+          description
+            "The geodetic system of the location data.";
+          leaf geodetic-datum {
+            type string {
+              pattern '[ -@\[-\^_-~]*';
+            }
+            default "wgs-84";
+            description
+              "A geodetic-datum defining the meaning of latitude,
+               longitude and height. The default is 'wgs-84' which is
+               used by the Global Positioning System (GPS). The value
+               SHOULD be comprised of all lower case ASCII characters
+               not including control characters (i.e., values 32..64,
+               and 91..126). The IANA registry further restricts the
+               value by converting all spaces (' ') to dashes ('-')";
+          }
+          leaf coord-accuracy {
+            type decimal64 {
+              fraction-digits 6;
+            }
+            description
+              "The accuracy of the latitude longitude pair. When
+               coord-accuracy is specified it overrides the
+               geodetic-datum implied accuracy. If Cartesian
+               coordinates are in use this accuracy corresponds to
+               the X and Y components";
+          }
+          leaf height-accuracy {
+            type decimal64 {
+              fraction-digits 6;
+            }
+            units "meters";
+            description
+              "The accuracy of height value. When specified it
+               overrides the geodetic-datum implied default. If
+               Cartesian coordinates ar in use this accuracy
+               corresponds to the Z component.";
+          }
+          // May wish to allow for height to be relative.
+          // If so need to decide if we have a boolean (to ground)
+          // or an enumeration (e.g., local ground, sea-floor,
+          // ground floor, containing object, ...) or even allow
+          // for a string for most generic but least portable
+          // comparable
+          // leaf height-relative {
+          // }
+        }
+      }
+      choice location {
+        mandatory true;
+        description
+          "The location data either in lat/long or Cartesian values";
+        case ellipsoid {
+          leaf latitude {
+            type degrees;
+            mandatory true;
+            description
+              "The latitude value on the astronomical body. The
+               definition and precision of this measurement is
+               indicated by the reference-frame value.";
+          }
+          leaf longitude {
+            type degrees;
+            mandatory true;
+            description
+              "The longitude value on the astronomical body. The
+               definition and precision of this measurement is
+               indicated by the reference-frame.";
+          }
+          leaf height {
+            type decimal64 {
+              fraction-digits 6;
+            }
+            units "meters";
+            description
+              "Height from a reference 0 value. The precision and '0'
+               value is defined by the reference-frame.";
+          }
+        }
+        case cartesian {
+          leaf x {
+            type decimal64 {
+              fraction-digits 6;
+            }
+            mandatory true;
+            description
+              "The X value as defined by the reference-frame.";
+          }
+          leaf y {
+            type decimal64 {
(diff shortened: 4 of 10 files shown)


jktjkt:
@EstherLerouzic , I would appreciate your input on the general approach.

Anyone who has comments on the set of parameters which are required, please comment as well.

Here's the tree rendering for reading convenience:
```
module: tip-photonic-equipment
  +--rw amplifier* [type]
  |  +--rw type                  string
  |  +--rw (simulation)
  |  |  +--:(edfa-fixed-gain)
  |  |  |  +--rw edfa-fixed-gain!
  |  |  |     +--rw gain    power
  |  |  |     +--rw nf      noise-figure
  |  |  +--:(edfa-variable-gain)
  |  |  |  +--rw edfa-variable-gain!
  |  |  |     +--rw nf-min               noise-figure
  |  |  |     +--rw nf-max               noise-figure
  |  |  |     +--rw gain-min             gain
  |  |  |     +--rw gain-flatmax         power
  |  |  |     +--rw gain-ripple*         db-ratio
  |  |  |     +--rw nf-ripple*           db-ratio
  |  |  |     +--rw dynamic-gain-tilt*   db-ratio
  |  |  +--:(edfa-advanced-model)
  |  |  |  +--rw edfa-advanced-model!
  |  |  |     +--rw gain-min                      gain
  |  |  |     +--rw gain-flatmax                  power
  |  |  |     +--rw gain-ripple*                  db-ratio
  |  |  |     +--rw nf-ripple*                    db-ratio
  |  |  |     +--rw dynamic-gain-tilt*            db-ratio
  |  |  |     +--rw nf-polynomial-coefficients
  |  |  |        +--rw a    polynomial-coefficient
  |  |  |        +--rw b    polynomial-coefficient
  |  |  |        +--rw c    polynomial-coefficient
  |  |  |        +--rw d    polynomial-coefficient
  |  |  +--:(edfa-openroadm)
  |  |  |  +--rw edfa-openroadm!
  |  |  |     +--rw nf-polynomial-coefficients
  |  |  |        +--rw a    polynomial-coefficient
  |  |  |        +--rw b    polynomial-coefficient
  |  |  |        +--rw c    polynomial-coefficient
  |  |  |        +--rw d    polynomial-coefficient
  |  |  +--:(edfa-dual-stage)
  |  |     +--rw edfa-dual-stage!
  |  |        +--rw preamp      -> /tip-photonic-equipment:amplifier/type
  |  |        +--rw booster     -> /tip-photonic-equipment:amplifier/type
  |  |        +--rw gain-min    power
  |  +--rw max-power-out         power
  |  +--rw frequency-min?        frequency <191.35>
  |  +--rw frequency-max?        frequency <196.1>
  |  +--rw autodesign-enabled?   boolean <false>
  |  +--rw autodesign-voa?       boolean <false>
  +--rw fiber* [type]
  |  +--rw type          string
  |  +--rw dispersion    decimal64
  |  +--rw gamma         decimal64
  +--rw transceiver* [type]
     +--rw type             string
     +--rw frequency-min?   frequency <191.35>
     +--rw frequency-max?   frequency <196.1>
     +--rw mode* [name]
        +--rw name             string
        +--rw baud-rate        uint64
        +--rw required-osnr    db-ratio
        +--rw tx-osnr          db-ratio
        +--rw grid-spacing     frequency
        +--rw tx-roll-off?     decimal64
        +--rw cost?            uint32 <1>
```
```
module: tip-photonic-topology
  +--rw elements* [uuid]
  |  +--rw uuid        string
  |  +--rw (element)
  |  |  +--:(amplifier)
  |  |  |  +--rw amplifier
  |  |  |     +--rw model          -> /tip-photonic-equipment:amplifier/tip-photonic-equipment:type
  |  |  |     +--rw gain-target    tip-photonic-equipment:gain
  |  |  |     +--rw tilt-target    tip-photonic-equipment:db-ratio
  |  |  +--:(fiber)
  |  |  |  +--rw fiber
  |  |  |     +--rw type      -> /tip-photonic-equipment:fiber/tip-photonic-equipment:type
  |  |  |     +--rw length?   decimal64
  |  |  +--:(concentrated-loss)
  |  |  |  +--rw concentrated-loss
  |  |  |     +--rw attenuation?   tip-photonic-equipment:db-ratio <0.0>
  |  |  +--:(transceiver)
  |  |     +--rw transceiver
  |  |        +--rw model    -> /tip-photonic-equipment:transceiver/tip-photonic-equipment:type
  |  +--rw location
  |     +--rw name?        string
  |     +--rw latitude?    geo-coordinate
  |     +--rw longitude?   geo-coordinate
  +--rw connections* [from to]
     +--rw from    -> ../../elements/uuid
     +--rw to      -> ../../elements/uuid
```

ojnas:
I'll have to look at the YANG models more in detail but in general I think this is the right approach.

If we're changing the input format anyway, I think it would be good if we could take #227 into account, i.e. separating the spectral model and the NF description in the amplifier model. Of course, this means some changes to the amplifier code as well.

About the topology model, I think it's at least worth to think about if we can augment the general ietf-network-topology model like they do in OpenROADM for example.

By "EDFA maximal gain", what parameter are you referring to? The `gain_flatmax` parameter is used in several places.

I think the roll-off follows the standard definition for raised-cosine pulse shaping, see e.g. the β parameter [here](https://en.wikipedia.org/wiki/Raised-cosine_filter). This parameter is not used anywhere in the current code but will be used in the new Raman feature, see PR #263.

jktjkt:
Thanks for looking into this.

> If we're changing the input format anyway, I think it would be good if we could take #227 into account, i.e. separating the spectral model and the NF description in the amplifier model. Of course, this means some changes to the amplifier code as well.

Sounds good. I'll comment there.

> About the topology model, I think it's at least worth to think about if we can augment the general ietf-network-topology model like they do in OpenROADM for example.

We talked about this last week with Esther, the outcome was to go with an ad-hoc model. At first, we wanted to use [CCAMP's model for impairment-aware optical network](https://github.com/younglee-ietf/ietf-optical-impairment-yang), but it is just not ready. I think we also looked at IETF models (the [Base Network model `ietf-network-model`](https://tools.ietf.org/html/rfc8345#section-4.1) and the [Base Network Topology model `ietf-network-topology`](https://tools.ietf.org/html/rfc8345#section-4.2)). They appear to focus on multi-layer adaptations. That's something we do not need now.

I think that these are good models in the long term. For example, it makes it possible, at least in theory, to use some common code for basic topology traversal. However, this will be quite invasive change already, and I wanted to make stuff as simple as possible in this "non-core" part. That said, I'm not really attached to ad-hoc topology, so I'll take a look at using `ietf-network-topology`. Further input is welcome.

> By "EDFA maximal gain", what parameter are you referring to? The `gain_flatmax` parameter is used in several places.

I'm referring to the maximal possible gain which can be set. My understanding is that this might be higher than `gain_flatmax` as it switches the amplifier into an extended range. I'll be hand-waving this explanation a bit, but I think that the code should track this so that the network designer can (perhaps deliberately, perhaps via auto-design?) create a network where the amp gain is higher than the `gain_flatmax`, yet still within the physical (or SW, I don't really know) capabilities of the amplifier.

> I think the roll-off follows the standard definition for raised-cosine pulse shaping, see e.g. the β parameter [here](https://en.wikipedia.org/wiki/Raised-cosine_filter). This parameter is not used anywhere in the current code but will be used in the new Raman feature, see PR #263.

Thanks. I was reading https://en.wikipedia.org/wiki/Roll-off and there it talks about decibels per octave. Adjusted in the model, thanks!

ojnas:
> I'm referring to the maximal possible gain which can be set. My understanding is that this might be higher than gain_flatmax as it switches the amplifier into an extended range.

I agree, see also discussion [here](https://github.com/Telecominfraproject/oopt-gnpy/issues/118#issuecomment-433953528). What's in the code currently is a `target_extended_gain` parameter configurable in the `Span` section in eqpt_config.json. It would probably make more sense to specify it per EDFA type_variety. Also, this parameter is currently only used for auto-design. There is no check that a manually set gain is within the extended range.

EstherLerouzic:
I have rapidly looked at the PR and comments and here are my first  comments. I will need a bit more time to check that we have no regression and only improvements with this change.
@jktjkt :
>  So instead of deferring to another JSON file for the NF-/gain-ripple/DGT, let's move everything in-line into the input data

I do not agree with this : the nf_ripple, gain_ripple and DGT are each a set of 96 points. I do not think it is a good idea to copy paste this in each amplifier block in the same file: it will not be human readable! Be aware that users create their own eqpt library based on vendors products, and this must be as simple as possible for them to build and to read (this is our usage).
Please keep the files separated !

> I somehow doubt that these data represent actual real-world results -- or do we really have a universal set of DGT for all amplifiers

not for all but for a given line product, the shape is similar. 
The advanced models enable user to input particular products lines if they have the measurements,
Jean-Luc proposed a std_medium_gain_advanced_config.json as an example and converted the data provided by @ggrammel  for Juniper amplifier. Again, I think this should not be put in the same file !

>the current code rejects dual-stage EDFAs built from single-coil fixed-gain modules (or doesn't it?)

You can build a dual stage with any type_varieties (except dual :) and especially fixed gain amp: this is how Raman amplification is handled right now (see hybrid_4_pumps_lowgain for example)  the 4pumps_raman is defined as a fixed gain amp. I think that the raman attribute is missing at this point in your description.

> create a network where the amp gain is higher than the gain_flatmax, yet still within the physical (or SW, I don't really know) capabilities of the amplifier

This is already the case: the autodesign build the network based on the extended gain range provided in the Span field. So the gain target can be higher than the max_flat_gain. but as @ojnas stated  
>There is no check that a manually set gain is within the extended range
Indeed there is no warning if the gain goes beyond extended gain range when the user just forces the amplifier type: this is an issue that I have been asked to solve by some users. I will post this as an issue (if one doesn't already exists in the issue list) 

on the topology: I can see that you have suppressed the operational and params containers. it simplifies, but maybe limits future possible differentiation between params requested by the user and params effectively possible with the tool.

This answer is a bit long (and long to write) , sorry for this, and maybe we can plan some confcall, to discuss live on this base ? In the mean time I will try to make proposal on this base as soon as I can

ggrammel:
I basically agree with Esther's ask to keep files separated, as the data would otherwise be too much. Few words about the description:
The data in the Juniper model is a representative figure i.e. it is a conservative average of many samples which inevitably vary in the manufacturing process. In particular we had the data measured only at a 100GHz grid, i.e. not 96 but 48 Wavelength.
We also have Lumentum data at our disposition in TIP. The data provided by Lumentum is a sample set of the factory acceptance test. So that data is not averaged and considered BOL. Also Lumentum used a 100GHz spacing to describe their modules.

Looking back few issues came up:
1) we would need a way to accept detailed data not only for 96ch but also for a subset of measured channels (e.g. 48ch). My take is that 48ch are pretty standard today. J.-L. apparently addressed this already in #221 
2) another issue at the time was how to calculate the polynomial from the data provided:
 <snip> your NF coefs are false because we need NF = f(gain_max-gain) and you gave NF = f(gain) <snip>
3) it is unclear how to aggregate ripple information. The NF ripple shapes were too variable to make good sense of. So we set the polynominal on the worst case wavelength and then set the ripple to be 0dB flat. 

in particular 2) may be a common mistake and I wonder if we can do something about falling into this trap.
it is not clear how 3) is supposed to work. Either the planner knows the ripple of a particular sample which may be the case in an online situation, or he doesn't and then what s the next best thing to do to cover the ripple?

jktjkt:
Thanks for taking a look.

> This answer is a bit long (and long to write) , sorry for this, and maybe we can plan some confcall, to discuss live on this base ? In the mean time I will try to make proposal on this base as soon as I can

Good idea. I'm available on Friday and on Monday. How does a morning (8am, 9am, 10am) sound? I think it's better to talk about this outside of the regular coders call (I think that Polito people would not be interested, for example).

> > So instead of deferring to another JSON file for the NF-/gain-ripple/DGT, let's move everything in-line into the input data
> 
> I do not agree with this : the nf_ripple, gain_ripple and DGT are each a set of 96 points. 

I *think* that the list might also be substantially shorter. The code assumes that the number of items in the list uniformly covers the whole supported spectrum as a piecewise linear function (at least that's my understanding of [`gnpy.core.elements.Edfa.interpol_params`](https://github.com/Telecominfraproject/oopt-gnpy/blob/8f735316f57c1ae5fae29c24cb839dfbb62e9922/gnpy/core/elements.py#L578) -- and I see now that Gert refers to this and that J-L implemented this recently). This means that for that common case where these parameters are zero, a simple `[0]` will do. But yeah, it's a long list :( when there are real measurements results.

> I do not think it is a good idea to copy paste this in each amplifier block in the same file: it will not be human readable! Be aware that users create their own eqpt library based on vendors products, and this must be as simple as possible for them to build and to read (this is our usage).

Basically, I *think* that each EDFA model will require a separate set of data. Is that correct? The data remains optional. If they are not provided, then the code will be using all zeros for all missing parameters.

I believe that the YANG models should really contain **everything** that is needed for using GNPy, and that's why I oppose an out-of-band file. In my opinion, that would complicate the (upcoming) API which will be based on YANG-modeled data. I'm mostly OK with the actual LUTs being "stashed away" into a separate model if that's what you prefer (but I think that it's one more indirection) -- it's just that I want it to be covered by (some) YANG as well.

What I want to achieve in the long run is an API for remote access to GNPy. Here's how I would like it to work:

- Start a GNPy server with no site-specific data. We could even run one as a demo service for our users.
- User opens a connection/session to that server, uploads the inventory and the network topology.
- Start running queries.

It is up for a debate whether to implement this via RESTCONF (plus some extension for per-session equivalent of NETCONF's `candidate` datastore), or via a set of REST APIs. We could even do the whole thing in a completely stateless manner, where each request contains the equipment library, the topology info, and some request specification (what services to compute, essentially).

If the detailed EDFA profile lives in a separate JSON file, then I think we're only using this machine-readable description language (YANG) half-way. There would have to be another, possibly out-of-band channel for uploading these data. How do we validate it? How is it uploaded? etc.

> > I somehow doubt that these data represent actual real-world results -- or do we really have a universal set of DGT for all amplifiers
> 
> not for all but for a given line product, the shape is similar.

Right, I meant "all different amplifier models". Perhaps I went a bit too far here :), you're right that we have three data sets in `examples/` already. Sorry. It's just that I think that it is not a good idea to use our existing `examples/default_edfa_config.json` by default for all amplifiers. Or is it?

> > the current code rejects dual-stage EDFAs built from single-coil fixed-gain modules (or doesn't it?)
> 
> You can build a dual stage with any type_varieties (except dual :) and especially fixed gain amp: this is how Raman amplification is handled right now (see hybrid_4_pumps_lowgain for example) the 4pumps_raman is defined as a fixed gain amp.

Ah, thanks. I got confused because the code looked at `gain_min` and `gain_flatmax`, and I thought that these are not initialized for the `fixed-gain` variety. I'll remove that TODO.

> I think that the raman attribute is missing at this point in your description.

I've deliberately ignored that for now because we do not have simulation code for Ramans. Or was that a mistake, and is there place for indicating "hey, this is a Raman amp" even now, before Alessio's code is merged?

> on the topology: I can see that you have suppressed the operational and params containers. it simplifies, but maybe limits future possible differentiation between params requested by the user and params effectively possible with the tool.

The interaction flow that I had in mind was:

- start the server
- send a self-contained request, including the inventory DB and the topology
- obtain back the topology with auto-design changes (etc) factored in. This is where GNPy can return adjusted gains, etc.

Perhaps I'm missing something here, so I'm open for comments (either here, or at the call if that is better for you). I don't really care about this part.

jktjkt:
Here are some notes from a VC with @EstherLerouzic from last Friday:

- Esther: it makes sense to decouple the DGT (etc) from the amplifier model definition in the equpiment library params because in the real world, people are relying on the default values (many amp models with the same LUTs)
- it is unclear whether a simple linear approximation with just two points `[2.7, 1.0]` would be "reasonable"
  - it fits with experiments conducted at Orange, which is a great data set

- Raman: been there since January
  - fixed gain amp with an excellent NF
  - no SRS simulation
  - therefore we need it in the YANG right now


- outside of the extended range: 0.8dB tilt for each 1dB of extra gain
  a new colleague from Orange, Bertrand Le Guyader, can comment on this

- let's get rid of the SI
  - transmission_main_example needs a default transponder type
  - unclear how to reconcile the f_min/max of the SI and TX

- a set of system parameters
  - autodesign needs to assume the full spectrum

ggrammel:
wondering what the previous comment had in mind related to:
"
  let's get rid of the SI
      transmission_main_example needs a default transponder type
      unclear how to reconcile the f_min/max of the SI and TX
"

the argument seems to pop up in this comment only and I have difficulties to follow what it means. in particular is "getting rid of SI" a general thought i.e. getting completely rid or a particular thought i.e. getting rid in some place of the code? Can you explain a little what is meant?

jktjkt:
Testing if a comment propagates to GerritHub -- please ignore me.

jktjkt:

jktjkt:
We revisited the DGT discussion on today's call, and I plotted the DGT values [from Orange (which are GNPy's defaults today)](https://github.com/Telecominfraproject/oopt-gnpy/blob/master/gnpy/example-data/default_edfa_config.json) and the [data from Juniper/Lumentum](https://github.com/Telecominfraproject/oopt-gnpy/blob/master/gnpy/example-data/Juniper-BoosterHG.json). The slopes appear to be very similar, but with a different Y scale:

### Orange
image
### Lumentum
image

PR #268, opened 2019-06-21
Update and rework dependency handling

jktjkt:
## requirements: only list what we really need

Listing all transitive dependencies, including their respective versions (as done by `pip freeze`, for example) is something which effectively leads to installing outdated and possibly vulnerable software.
    
Let's get rid of the transitive dependencies.
    
Thanks to Esther for noticing.

## Update dependencies
Pinning dependencies to a specific version is safe in terms of preventing accidental breakage, but it has a cost of possibly using outdated dependencies.
    
Let's see if we can rely on the semantic versioning of our dependencies here.
    
I'm doing this instead of other approaches (such as splitting the top-level deps from "the rest of the dep chain" [1][2]) because I require additional tools in my venv, such as the python-lnaguage-server. There would be little point in injecting these into requirements to be used by other people.
    
[1] https://www.kennethreitz.org/essays/a-better-pip-workflow
[2] https://github.com/jktjkt/oopt-gnpy/commit/59eb51c0268efe74d7cbaeb8fff86b3226f82a99


EstherLerouzic:
:+1

Issue #272, opened 2019-06-21
path constraint Excel entry does not support well non roadm names


EstherLerouzic:
Autodesign naming of elements adds new syntax in names which makes difficult for user to give ILA names as constraints. Add a module that recover the correct information for the constraints.

EstherLerouzic:
Solved

PR #278, opened 2019-07-03
Ila names in constraints

EstherLerouzic:
Enable ILA and FUSED type in routing constraints with xls input
    
    - builds correspondance dicts between input name from excel
      and names created with convert.py and autodesign in network.py
    - correct the corresp_name dicts according to the effective
      network autodesign. This supports the case of fiber splitting
      and of fused elements
    - include the case of parallel links with only one hop
    - interpret the node list constraint given by the user with the dict
    - filter the constraints that are not applicable
    
    ILA and FUSED constraints must be filled with the next node
    information in order to avoid confusion on the direction.
    for example
         eg   
'
 a----b-----c
 |       |       |
 i       j       k
 |       |       |
 e----f-----g
 '   
a constraint 'j' given for service i to k leads to 2 possible direction:
    i-a-b-j-f-g-k
    i-e-f-j-b-c-k
    the user must indicate the chosen direction. This ambiguity does not
    exist with network input in json format (names are unique).

This PR adresses #272     
TODO: 
complete readme files with info
use this to correct plot

### DIFF
--- a/examples/path_requests_run.py
+++ b/examples/path_requests_run.py
@@ -19,11 +19,12 @@ from collections import namedtuple
 from logging import getLogger, basicConfig, CRITICAL, DEBUG, INFO
 from json import dumps, loads
 from numpy import mean
+from gnpy.core.convert import corresp_names
 from gnpy.core.service_sheet import convert_service_sheet, Request_element, Element
 from gnpy.core.utils import load_json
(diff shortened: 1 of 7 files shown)


EstherLerouzic:
close this one and will rebase on develop

PR #289, opened 2019-09-01
Raman: linear interpolation of channel NLI

### LINKED ISSUE #288: problems in interaction between RamanFiber and Edfa elements
The RamanFiber only calculates NLI for channels specified in "raman_computed_channels" in sim_params.json. For all other channels in the SpectralInformation it sets NLI to `nan`. This creates problems when the SpectralInformation is passed on to an Edfa element which tries to calculate total power in each carrier:

`pin = array([c.power.signal+c.power.nli+c.power.ase for c in carriers])`

resulting in `nan` for most carriers as well as for the calculated total input and output powers for the whole spectrum. This can be seen for example running the main example:

`python3 transmission_main_example.py raman_edfa_example_network.json --sim-params sim_params.json`

which outputs `nan` for Edfa Power In and Power Out as well as for SNR total. But it may also result in other more subtle errors besides this output.

jktjkt:
The Raman engine computes NLI just for a subset of channels; this is an important speed optimization because the computation is rather CPU heavy. However, the code just left NaNs in place for NLI of those channels which were not explicitly simulated.

This is wrong because these NaNs propagate all the way to the total input/output powers per the whole spectrum, etc, leading to NaNs being shown to the user.

This patch uses a very simple linear approximation just in order to prevent these NaNs.

fixes #288
Cc: @aleFerrari

### DIFF
--- a/examples/transmission_main_example.py
+++ b/examples/transmission_main_example.py
@@ -18,7 +18,7 @@ from pathlib import Path
 from json import loads
 from collections import Counter
 from logging import getLogger, basicConfig, INFO, ERROR, DEBUG
-from numpy import linspace, mean, log10, isnan
+from numpy import linspace, mean, log10
 from matplotlib.pyplot import show, axis, figure, title, text
 from networkx import (draw_networkx_nodes, draw_networkx_edges,
                       draw_networkx_labels, dijkstra_path)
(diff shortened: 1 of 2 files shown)


ojnas:
Looks like a neat and simple solution to the issue. Note that with this fix, running the transmission main example with `--show-channels` shows SNR for all channels, not just for the channels in `sim_params.raman_computed_channels`. If that is not desired, transmission_main_example.py should also be changed (as a minumum the check `if not isnan` can be removed.

jktjkt:
> Looks like a neat and simple solution to the issue. Note that with this fix, running the transmission main example with `--show-channels` shows SNR for all channels, not just for the channels in `sim_params.raman_computed_channels`. If that is not desired, transmission_main_example.py should also be changed (as a minumum the check `if not isnan` can be removed.

Thanks for a review and for catching this. I *think* that it makes sense to show all channels here. I tried a change which adds another column for simulation/approximation as the source of the NLI estimate, but I think it only clutters the output.

jktjkt:
(The latest force-push is to fix some LGTM warnings. It's a nifty tool!)

PR #291, opened 2019-09-02
CI: tests: include examples in coverage reports

jktjkt:
This might be a wee bit controversial. On one hand, having "real", fine-grained tests and the associated coverage reports is very valuable in identifying code which is suspicious because there's nothing guarding that it still works.

On the other hand, the recent Raman work came with no test coverage.

I think that it's a reasonable compromise to introduce the existing examples into the autotest suite. That way we can see whether changes at least touch stuff that is "needed by the examples"...


jktjkt:
This does not do the job; the `subproccess` environment apparently does not preserve the `codecov` options.

PR #293, opened 2019-09-03
Update SNR before printing Transceiver output

### LINKED ISSUE #262: output from transmission_main_example.py
Running `python3 transmission_main_example.py` results in the following output:

```
Transceiver Site_B
  OSNR ASE (0.1nm, dB):      34.34
  OSNR ASE (signal bw, dB):  30.26
  SNR total (signal bw, dB): 27.70
  SNR total (0.1nm, dB): 31.78

Transmission result for input power = 0.00dBm :
Transceiver Site_B
  OSNR ASE (0.1nm, dB):      32.03
  OSNR ASE (signal bw, dB):  27.95
  SNR total (signal bw, dB): 26.27
  SNR total (0.1nm, dB): 30.35
```

It is confusing to see different SNR values printed for the same `Transceiver Site_B`. The reason this happens is that the SNRs are updated to include `tx_osnr` and `add_drop_osnr` after the last element of the path is printed (see the `propagate2` function in `request.py`).

I think the printed SNRs for the destination transceiver should already include all contributions. In this case we could also avoid printing the same output twice as proposed in PR #261 by @jktjkt .

ojnas:
This modifies the propagate functions in request.py so that the update_snr function of the last transceiver is called before printing the transceiver output (when show=True) in order to show the correct SNR including tx_osnr and add_drop_osnr contributions. I could not come up with a better way than to split the last element out from the for loop and repeating some code.

Fixes #262

### DIFF
--- a/gnpy/core/request.py
+++ b/gnpy/core/request.py
@@ -397,11 +397,14 @@ def propagate(path, req, equipment, show=False):
     si = create_input_spectral_information(
         req.f_min, req.f_max, req.roll_off, req.baud_rate,
         req.power, req.spacing)
-    for el in path:
+    for el in path[:-1]:
         si = el(si)
         if show :
             print(el)
+    si = path[-1](si)
     path[-1].update_snr(req.tx_osnr, equipment['Roadm']['default'].add_drop_osnr)
+    if show :
+        print(path[-1])
     return path
 
 def propagate2(path, req, equipment, show=False):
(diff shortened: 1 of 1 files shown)


EstherLerouzic:
thanks for the change ! I agree with it.
The alternative way for the propagate functions would have been to test in the loop if "the element is the last and an instance of Transceiver". But this does not work for the propagate_and_optimize_mode function.

ojnas:
> thanks for the change ! I agree with it.

Thanks!

> The alternative way for the propagate functions would have been to test in the loop if "the element is the last and an instance of Transceiver".

Yeah, I thought of that alternative but it felt kind of silly to have a test for every iteration in the loop that you already know will only be true for the last element.

jktjkt:
Thanks for this, and sorry for a delay in responding. I think that it is easier to just move the path printing outside of the loop, which is what I did in commit acafc78 in #299.

PR #295, opened 2019-09-17
Automatic building of Docker images

### LINKED ISSUE #260: Gnpy docker

jktjkt:
A third, independent job in Travis CI just for building a container
image (and testing it a little bit). Here's how to use it once it is
built and published and pulled:

```shell-session
  $ docker run -it --rm --volume $(pwd):/shared telecominfraproject/oopt-gnpy
```

One can also pass commands to it directly. Note that the *relative* path specifier given as `./` is required. One could possibly get around it by `$PATH` manipulation, but hey, I have to stop somewhere.

```shell-session
  $ docker run -it --rm --volume $(pwd):/shared telecominfraproject/oopt-gnpy ./transmission_main_example.py
```

Thanks to @ojnas for the `--volume` option trick and for that early non-overwriting copying for the `example/` directory.

The `install` phase in Travis CI is apparently common for all jobs in the build matrix (it's rather confusing what they call a "stage" and what is a "job" for them, and what their mutual relation is). Because there's no point in doing any kind of installation in case when we're building for Docker, let's just move the old `install` block into `script`. With just that, however, Travis adds an implicit `pip install -r requirements.txt` in an attempt at being helpful. In short, having completely different "stuff to be done as a check" is very painful with Travis.

The individual "script lines" in a job's `script` section are always executed, all of them, even if a previous one failed. That's why I moved the actual Docker invocation into an external shell script. When they were not in a shell script and the script lines contained `set -e`, then a failure of a particular command would cause the build to be marked as "errored" instead of "failed" within Travis. Also, the multiline if conditionals were rather painful to write.

One commit might end up in different branches. If that happens, the second build would overwrite the image tag in the docker registry, which is rather suboptimal. Let's try to fetch that image first, and only update the latest/stable tags if the image was already available beforehand.

Fixes #260

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM python:3.7-slim
+COPY . /oopt-gnpy
+WORKDIR /oopt-gnpy
+RUN python setup.py install
+WORKDIR /shared/examples
+ENTRYPOINT ["/oopt-gnpy/.docker-entry.sh"]
+CMD "/bin/bash"
--- a/.docker-entry.sh
+++ b/.docker-entry.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+cp -nr /oopt-gnpy/examples /shared
+exec "$@"
--- a/.docker-travis.sh
+++ b/.docker-travis.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+set -e
+
+IMAGE_NAME=telecominfraproject/oopt-gnpy
+IMAGE_TAG=$(git describe --tags)
+
+ALREADY_FOUND=0
+docker pull ${IMAGE_NAME}:${IMAGE_TAG} && ALREADY_FOUND=1
+
+if [[ $ALREADY_FOUND == 0 ]]; then
+  docker build . -t ${IMAGE_NAME}
+  docker tag ${IMAGE_NAME} ${IMAGE_NAME}:${IMAGE_TAG}
+
+  # shared directory setup: do not clobber the real data
+  mkdir trash
+  cd trash
+  docker run -it --rm --volume $(pwd):/shared ${IMAGE_NAME} ./transmission_main_example.py
+else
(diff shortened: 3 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM python:3.7-slim
+COPY . /oopt-gnpy
+WORKDIR /oopt-gnpy
+RUN python setup.py install
+WORKDIR /shared/examples
+ENTRYPOINT ["/oopt-gnpy/.docker-entry.sh"]
+CMD "/bin/bash"
```
ojnas: Minor detail perhaps but is there a reason you're not using the preferred _exec_ form of CMD, i.e. `CMD ["/bin/bash"]`?

jktjkt: Thanks, #297

PR #261, opened 2019-06-17
Prettyfication of transmission_main_example

jktjkt:
Esther suggested to create a small animation showing how the tool works in action. Creating a simple screencast led me to make the UI a bit easier to follow. I'm open to suggestions on what to highlight, what to remove, etc., but I think that this PR represents a positive change overall :)

Issue #246, opened 2019-05-27
Add a show option on path_requests_run.py 


EstherLerouzic:
When launching path_requests_run.py no detailed propagation is provided on the different computed paths. User needs to run separate transmission_main program to see the detail of the propagation, however the path might be different (constrained in path_requests_run and shortest path in transmission_main_example.py).

A 'show' tag on each line in the excel input service sheet could help to see the propagation details. The option might be reserved for excel input only.

Additionally enable constraints (include or exclude nodes) on transmission_main_example.py might be useful

ojnas:
Just FYI, an implementation of k-shortest-paths and exclude nodes constraints have been proposed here:

https://review.gerrithub.io/c/Telecominfraproject/oopt-gnpy/+/494445
https://review.gerrithub.io/c/Telecominfraproject/oopt-gnpy/+/494458

@ggrammel, exactly the use case you described is supported  at https://gnpy.app/.

ggrammel:
Thanks @ojnas, this works quite nicely!
One thing I observed when testing was that the Graph is not always corresponding to the simulation. (see screenshot attached). If it is some bug in in the library, Perhaps a "refresh" buttow would do:
<img width="1236" alt="Screen Shot 2020-06-12 at 12 29 36" src="https://user-images.githubusercontent.com/15140692/84493837-c2266200-aca8-11ea-895f-4daf09ec76e4.png">

ojnas:
@ggrammel, that is strange. Do you mean that the graph and results are not updated when you run a new simulation? I can't reproduce this. Could you please describe the steps you have performed when this happens?

ggrammel:
Re-checked and seems a Browser issue (see below)

**Original sequence (firefox 77.0.1):**
1. pick the CORONET_CONUS topology 
2. select k=4
3. calculate path Atlanta Scranton ==> here I observed a strange routing around New-York area that might be correct, but looks a bit funny for the path-2/3 (if I remember well)
4. to analyze the routing I zoomed into the new-york area after selecting each path (k=1..4)
4. because zooming is a bit tricky (OK to zoom in, but couldn't figure out zooming out again in  a controlled way), I decided to re run the scenario directly between New-York and Scranton
5. changed the source to New York
6. keeping k = 4
7. Running the simulation
... and here we are.

I could reproduce a similar but more subtle case right now. Running through the same steps as above. then
[...]

PR #299, opened 2019-09-19
Refactor printing

### LINKED ISSUE #262: output from transmission_main_example.py

jktjkt:
(This replaces #261 by me and #293 by Jonas.)

##     Remove debug printing from propagate()
This is inspired by #293. The original issue was that the transponder OSNR was not accounted for. Rather than making the `propagate()` and `propagate2()` more complex, let's move the actual path printing to the demo code. There's no secret sauce in there, it's just a simple for-each-print thingy, so let's remove it from the library code.
    
fixes #262

## doc: fiber length summary in km, not meters

Reading `80000m` is a bit more complex than just `80 km`. Also let's add a space between the numebr and the unit for better readability.

## examples: color highlighting of the "most interesting output"
    
I think that this SNR value represents the most important output of this  example. There's plenty of debugging info on display already, so let's make this one more prominent.
    
I was thinking about moving the highlighting to elements' each `__str__()` function, but that felt a bit like layering violation to me. I could have also changed just the transponder's `__str__()`. In the end, I think that repeating just the final GSNR at the link-end transponder makes most sense here. This is aligned with what we talked about at yesterday's (on 2019-09-18 -- note that this is a backport from #261) demo for Microsoft, after all.

### DIFF
--- a/examples/transmission_main_example.py
+++ b/examples/transmission_main_example.py
@@ -161,7 +161,7 @@ def main(network, equipment, source, destination, sim_params, req=None):
             print(f'\nTransmission result for input power = {lin2db(req.power*1e3):.2f} dBm:')
         else:
             print(f'\nTransmission results:')
-        print(f'  Final GSNR (signal bw): {ansi_escapes.cyan}{mean(destination.snr):.02f} dB{ansi_escapes.reset}')
+        print(f'  Final SNR total (signal bw): {ansi_escapes.cyan}{mean(destination.snr):.02f} dB{ansi_escapes.reset}')
 
         #print(f'\n !!!!!!!!!!!!!!!!!     TEST POINT         !!!!!!!!!!!!!!!!!!!!!')
         #print(f'carriers ase output of {path[1]} =\n {list(path[1].carriers("out", "nli"))}')
--- a/examples/transmission_main_example.py
+++ b/examples/transmission_main_example.py
@@ -153,7 +153,10 @@ def main(network, equipment, source, destination, sim_params, req=None):
             print(f'\nPropagating with input power = {lin2db(req.power*1e3):.2f}dBm :')
         else:
             print(f'\nPropagating in gain mode: power cannot be set manually')
-        infos = propagate2(path, req, equipment, show=len(power_range)==1)
+        infos = propagate2(path, req, equipment)
+        if len(power_range) == 1:
+            for elem in path[:-1]:
+                print(elem)
         if power_mode:
             print(f'\nTransmission result for input power = {lin2db(req.power*1e3):.2f}dBm :')
         else:
--- a/examples/transmission_main_example.py
+++ b/examples/transmission_main_example.py
@@ -150,19 +150,18 @@ def main(network, equipment, source, destination, sim_params, req=None):
     for dp_db in power_range:
         req.power = db2lin(pref_ch_db + dp_db)*1e-3
         if power_mode:
-            print(f'\nPropagating with input power = {lin2db(req.power*1e3):.2f}dBm :')
+            print(f'\nPropagating with input power = {ansi_escapes.cyan}{lin2db(req.power*1e3):.2f} dBm{ansi_escapes.reset}:')
         else:
-            print(f'\nPropagating in gain mode: power cannot be set manually')
+            print(f'\nPropagating in {ansi_escapes.cyan}gain mode{ansi_escapes.reset}: power cannot be set manually')
         infos = propagate2(path, req, equipment)
         if len(power_range) == 1:
-            for elem in path[:-1]:
+            for elem in path:
                 print(elem)
         if power_mode:
-            print(f'\nTransmission result for input power = {lin2db(req.power*1e3):.2f}dBm :')
+            print(f'\nTransmission result for input power = {lin2db(req.power*1e3):.2f} dBm:')
         else:
             print(f'\nTransmission results:')
-        #info message in gain mode
-        print(destination)
+        print(f'  Final GSNR (signal bw): {ansi_escapes.cyan}{mean(destination.snr):.02f} dB{ansi_escapes.reset}')
 
         #print(f'\n !!!!!!!!!!!!!!!!!     TEST POINT         !!!!!!!!!!!!!!!!!!!!!')
         #print(f'carriers ase output of {path[1]} =\n {list(path[1].carriers("out", "nli"))}')
@@ -183,8 +182,7 @@ def main(network, equipment, source, destination, sim_params, req=None):
                         'OSNR_ASE_signal_bw'    : round(mean(destination.osnr_ase),2),
                         'SNR_nli_signal_bw'     : round(mean(destination.osnr_nli),2),
                         'SNR_total_signal_bw'   : round(mean(destination.snr),2)
-                                })          
-        #info message in gain mode
+                                })
     write_csv(result_dicts, 'simulation_result.csv')
     return path, infos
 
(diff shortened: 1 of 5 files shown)


EstherLerouzic:
I do not agree with this removing of 'show' argument in the functions: this is a feature that we need, not fo pretty reason but for design analysis. it is being used by users.

gnpy/core/request.py
```diff
@@ -393,18 +393,16 @@ def compute_constrained_path(network, req):
 
     return total_path
 
-def propagate(path, req, equipment, show=False):
+def propagate(path, req, equipment):
```
EstherLerouzic: please leave the show argument !

jktjkt:
> I do not agree with this removing of 'show' argument in the functions: this is a feature that we need, not fo pretty reason but for design analysis. it is being used by users.

Hi Esther, can you please clarify what feature you have in mind? The output of the example code remains effectively the same. [This patch](https://github.com/Telecominfraproject/oopt-gnpy/pull/299/commits/acafc78456a2701cfed210a2d3037cc6a9e6e0a3#diff-25465525343a2233183dc73986d8af2fL156) is simply moving these two lines out of the `propagate2()` function:

```python
for element in path:
  print element
```

The functionality is preserved. The only callers of these functions which passed a `show=True` were in `./examples/transmission_main_example.py`, and path printing is preserved in there.

ggrammel:
Not sure if it applies here:
doc: fiber length summary in km, not meters
we decided early on that all parameters we are going to uses have to be expressed in SI base units. i.e. mks and dB, Hz. This may not look pretty but helps integrating code as there is no doubt about in which units things are expressed.

jktjkt:
> we decided early on that all parameters we are going to uses have to be expressed in SI base units. i.e. mks and dB, Hz. This may not look pretty but helps integrating code as there is no doubt about in which units things are expressed.

This change applies purely to the user-facing output when running examples. Previously, it would print out the total length in meters, which is, in my opinion, rather unreadable. The actual input parameters and the simulation code remains unchanged.

Do you want me to remove this part of the changes nonetheless?

EstherLerouzic:
Me and other users from orange also use the show function with path requests run. I did not have time to propose code for this here, but you can refer to this issue #246 where I specified it. propagating the show option in the different function is very useful for my studies: I can use it too print specific information when I am in the functions. So do not see advantages in suppressing it, since I will need to add another way to have it.

ggrammel:
For presentation that’s OK. The key is to keep consistency in the code. Presentation to users gain from using “pretty” numbers.

jktjkt:
> Me and other users from orange also use the show function with path requests run. I did not have time to propose code for this here, but you can refer to this issue #246 where I specified it. propagating the show option in the different function is very useful for my studies: I can use it too print specific information when I am in the functions. So do not see advantages in suppressing it, since I will need to add another way to have it.

Thanks, I of course did not know how your internal code is using these functions. Would you mind sharing a typical script that you're using so that I have a better idea, and that I could perhaps help prepare a patch which does not cause you much trouble? I have a feeling that adding a one-liner for-each just after the call to `propagate_XXX` should not be that much trouble, but perhaps I'm missing something.

### Why am I changing these

In general, I stand by my point that calling `print()` from inside library functions is a pattern that we should get a rid of. If this is about providing nice interface to the users, then let's make this explicit -- return data from the library, and let the *user-facing code* control the presentation. if this is about debugging, let's use proper tools for this, i.e., the `logging` module.

I am not touching these functions just for sake of doing changes. There was a real bug in there (#262) that the output was wrong. Fixing this bug (see the patch given in #293) made the code more spaghetti. The real root cause of this bug is that printing happened "too early", and therefore it is technically a better solution to walk the path first, perform the computation, and only then print the result. Which begs the question "why not move the printing outside of the propagation function", in my professional opinion.

But clearly you have some code that I have not seen yet, so let's take a look at it and find out if we can make **both** of these pieces better :).

EstherLerouzic:
I understand your point.
Here are the needs:
In path_requests_run.py the function compute_path_with_disjunction calls propagate. When I need specific analysis I just set show to True and put the stdout in a text file, for analysis. My plan is to have this show option per request adding an option in service file and have the result saved in a specific file (automatically not using > res.text :)). No code ready, only plan.
Some colleagues have gone a step further (sorry I do not have the code ready yet), and show more details than element string has (getting the attributes directly). For now it is added directly on the path_request_run: after having all paths propagated they loop again on all paths, create the string they need for each element of each request and print it in a separate file). Again I think that a show option could ease.

This cases are for design analysis (not debugging)

For debugging I think that simple prints  are enough as you said.

examples/transmission_main_example.py
```diff
@@ -150,16 +150,18 @@ def main(network, equipment, source, destination, sim_params, req=None):
...
+            for elem in path:
+                print(elem)
         if power_mode:
-            print(f'\nTransmission result for input power = {lin2db(req.power*1e3):.2f}dBm :')
+            print(f'\nTransmission result for input power = {lin2db(req.power*1e3):.2f} dBm:')
         else:
             print(f'\nTransmission results:')
-        #info message in gain mode
-        print(destination)
+        print(f'  Final GSNR (signal bw): {ansi_escapes.cyan}{mean(destination.snr):.02f} dB{ansi_escapes.reset}')
```
ojnas: The term `GSNR` is not used anywhere else in the code and is not a well established term as far as I know. In the transceiver output printing, `SNR total` is used. Let's at least use the same term in both outputs to avoid confusion.

jktjkt: Thanks, I've added a commit which implements just that. In the longer term, I would like to switch to "GSNR" as this is what the Polito team has been using in their presentations for some time, but you're right that it is basically undefined right now.

jktjkt:
@EstherLerouzic , you can get an equivalent output by adding explicit printing into the example code. It is a trivial one-liner:

```diff
diff --git a/examples/path_requests_run.py b/examples/path_requests_run.py
index a5fc979..ddfb247 100755
--- a/examples/path_requests_run.py
+++ b/examples/path_requests_run.py
@@ -167,6 +167,7 @@ def compute_path_with_disjunction(network, equipment, pathreqlist, pathlist):
         if total_path :
             if pathreq.baud_rate is not None:
                 total_path = propagate(total_path,pathreq,equipment)
+                for el in total_path: print(el)
                 temp_snr01nm = round(mean(total_path[-1].snr+lin2db(pathreq.baud_rate/(12.5e9))),2)
                 if temp_snr01nm < pathreq.OSNR :
                     msg = f'\tWarning! Request {pathreq.request_id} computed path from {pathreq.source} to {pathreq.destination} does not pass with {pathreq.tsp_mode}\n' +\
```
I've pushed a commit (ec9eb8d) which adds this (as a commented out code). I am not happy with this result, but it's better than being stuck on a minor disagreement.

PR #303, opened 2019-09-23
Test on spectrum assignment

EstherLerouzic:
** Adding a set of tests on spectrum_assignment functions **
This PR completes capacity_planning-part2 with tests on spectrum assignment functions.
It includes both unitary test and functional tests.

- Test oms building
- Test the alignment function    
- Test that the assignment of n and m values is correct
- Test assignment with reversed path
- Test with a path configuration
- Test assignment limits
- Check that spectrum assignment of 13,7 is correct in Hz

### DIFF
--- a/tests/test_spectrum_assignment.py
+++ b/tests/test_spectrum_assignment.py
@@ -1,0 +1,378 @@
+#!/usr/bin/env python3
+# TelecomInfraProject/gnpy/examples
+# Module name: test_spectrum_assignment.py
+# Version:
+# License: BSD 3-Clause Licence
+# Copyright (c) 2018, Telecom Infra Project
+
+"""
+@author: esther.lerouzic
+
+"""
+
+from pathlib import Path
+from copy import deepcopy
+from json import loads
+from math import ceil
+import pytest
+from gnpy.core.equipment import load_equipment, automatic_nch
+from gnpy.core.network import load_network, build_network
+from gnpy.core.utils import lin2db
+from gnpy.core.elements import Roadm, Transceiver
+from gnpy.core.spectrum_assignment import (build_oms_list, align_grids, nvalue_to_frequency,
+                                           bitmap_sum, m_to_freq, slots_to_m, frequency_to_n,
+                                           Bitmap, spectrum_selection, pth_assign_spectrum)
+from gnpy.core.exceptions import SpectrumError
+from gnpy.core.request import compute_path_dsjctn, find_reversed_path
+from examples.path_requests_run import (requests_from_json, disjunctions_from_json,
+                                        correct_disjn, compute_path_with_disjunction)
+
+TEST_DIR = Path(__file__).parent
+DATA_DIR = TEST_DIR / 'data'
+EQPT_FILENAME = DATA_DIR / 'eqpt_config.json'
+NETWORK_FILENAME = DATA_DIR / 'testTopology_auto_design_expected.json'
+SERVICE_FILENAME = DATA_DIR / 'testTopology_services_expected.json'
+
+@pytest.fixture()
+def eqpt():
+    """ common setup for tests: builds network, equipment and oms only once
+    """
+    equipment = load_equipment(EQPT_FILENAME)
+    return equipment
+
+@pytest.fixture()
+def setup(eqpt):
+    """ common setup for tests: builds network, equipment and oms only once
+    """
+    equipment = eqpt
+    # fix band to be independant of changes in json file
+    equipment['SI']['default'].f_min = 191300000000000.0
+    equipment['SI']['default'].f_max = 196100000000000.0
+    network = load_network(NETWORK_FILENAME, equipment)
+    p_db = equipment['SI']['default'].power_dbm
+    p_total_db = p_db + lin2db(automatic_nch(equipment['SI']['default'].f_min,\
+        equipment['SI']['default'].f_max, equipment['SI']['default'].spacing))
+    build_network(network, equipment, p_db, p_total_db)
+    oms_list = build_oms_list(network, equipment)
+    return network, oms_list
+
+def test_oms(setup):
+    """ tests that the oms is between two roadms, that there is no roadm or transceivers in the oms
+        except end points, checks that the id of oms is present in the element and that the element
+        oms id is consistant
+    """
+    network, oms_list = setup
+    for oms in oms_list:
+        assert isinstance(oms.el_list[0], Roadm) and isinstance(oms.el_list[-1], Roadm)
+        for i, elem in enumerate(oms.el_list[1:-2]):
+            assert not isinstance(elem, Roadm)
+            assert elem  in network.nodes()
+            assert elem.oms.oms_id == oms.oms_id
+            print(f'expected {elem.uid}, obtained {oms.el_id_list[i+1]}')
+            assert elem.uid == oms.el_id_list[i+1]
+
+@pytest.mark.parametrize('nmin', [-288, -260, -300])
+@pytest.mark.parametrize('nmax', [480, 320, 450])
+def test_aligned(nmin, nmax, setup):
+    """ checks that the oms grid is correctly aligned. Note that bitmap index uses guardband
+        on both ends so that if nmin, nmax = -200, +200, min/max navalue in bitmap are
+        -224, +223, which makes 223 -(-224) +1 frequencies
+    """
+    network, oms_list = setup
+    # f_min = 193.1e12 - 280 * 0.00625e12
+    # f_max = 193.1e12 + 320 * 0.00625e12
+    # f_min = 1931.1 f_max = 195.1
+    grid = 0.00625e12
+    guardband = 0.15e12
+    nguard = 24
+    freq_min = 193.1e12 + nmin * 0.00625e12
+    freq_max = 193.1e12 + nmax * 0.00625e12
+    print('initial spectrum')
+    print(nvalue_to_frequency(oms_list[10].spectrum_bitmap.freq_index_min) * 1e-12,
+          nvalue_to_frequency(oms_list[10].spectrum_bitmap.freq_index_max) * 1e-12)
+    # checks initial values consistancy
+    ind_max = len(oms_list[10].spectrum_bitmap.bitmap) - 1
+    print('with guardband', oms_list[10].spectrum_bitmap.getn(0),
+          oms_list[10].spectrum_bitmap.getn(ind_max))
+    print('without guardband', oms_list[10].spectrum_bitmap.freq_index_min,
+          oms_list[10].spectrum_bitmap.freq_index_max)
+    nvalmin = oms_list[10].spectrum_bitmap.getn(0) + nguard
+    nvalmax = oms_list[10].spectrum_bitmap.getn(ind_max) - nguard + 1
+
+    # min index in bitmap must be consistant with min freq attribute
+    print(f'test1 expected: {nvalmin}')
+    print(f'freq_index_min: {oms_list[10].spectrum_bitmap.freq_index_min},')
+    assert nvalmin == oms_list[10].spectrum_bitmap.freq_index_min
+    print(f'test2 expected: {nvalmax}')
+    print(f'freq_index_max: {oms_list[10].spectrum_bitmap.freq_index_max}')
+    assert nvalmax == oms_list[10].spectrum_bitmap.freq_index_max
+    oms_list[10].update_spectrum(freq_min, freq_max, grid=grid, guardband=guardband)
+    # checks that changes are applied on bitmap and freq attributes of Bitmap object
+    print(f'expected: {nmin}, {nmax}')
+    print(f'test3 obtained: {oms_list[10].spectrum_bitmap.freq_index_min},' +\
+          f' {oms_list[10].spectrum_bitmap.freq_index_max}')
+    assert (nmin == oms_list[10].spectrum_bitmap.freq_index_min and
+            nmax == oms_list[10].spectrum_bitmap.freq_index_max)
+
+    print('novel spectrum')
+    print(nvalue_to_frequency(oms_list[10].spectrum_bitmap.freq_index_min) * 1e-12,
+          nvalue_to_frequency(oms_list[10].spectrum_bitmap.freq_index_max) * 1e-12)
+    ind_max = len(oms_list[10].spectrum_bitmap.bitmap) - 1
+    print('with guardband', oms_list[10].spectrum_bitmap.getn(0),
+          oms_list[10].spectrum_bitmap.getn(ind_max))
+    print('without guardband', oms_list[10].spectrum_bitmap.freq_index_min,
+          oms_list[10].spectrum_bitmap.freq_index_max)
+    nvalmin = oms_list[10].spectrum_bitmap.getn(0) + nguard
+    nvalmax = oms_list[10].spectrum_bitmap.getn(ind_max) - nguard + 1
+    # min index in bitmap must be consistant with min freq attribute
+    print(f'expected: {nvalmin}')
+    print(f'freq_index_min: {oms_list[10].spectrum_bitmap.freq_index_min},')
+    print('inconsistancy in Bitmap object')
+    assert nvalmin == oms_list[10].spectrum_bitmap.freq_index_min
+    print(f'expected: {nvalmax}')
+    print(f'freq_index_max: {oms_list[10].spectrum_bitmap.freq_index_max}')
+    print('inconsistancy in Bitmap object')
+    assert nvalmax == oms_list[10].spectrum_bitmap.freq_index_max
+    oms_list = align_grids(oms_list)
+    ind_max = len(oms_list[10].spectrum_bitmap.bitmap) - 1
+    nvalmin = oms_list[10].spectrum_bitmap.getn(0)
+    nvalmax = oms_list[10].spectrum_bitmap.getn(ind_max)
+    print(f'expected: {min(nmin, nvalmin)}, {max(nmax, nvalmax)}')
+    print(f'expected: {nmin, nmax}')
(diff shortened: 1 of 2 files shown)


EstherLerouzic:
@jktjkt : could you please have a look on this PR and tell me how to improve ?

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+    """
+    network, oms_list = setup
+    # f_min = 193.1e12 - 280 * 0.00625e12
+    # f_max = 193.1e12 + 320 * 0.00625e12
+    # f_min = 1931.1 f_max = 195.1
+    grid = 0.00625e12
+    guardband = 0.15e12
+    nguard = 24
+    freq_min = 193.1e12 + nmin * 0.00625e12
+    freq_max = 193.1e12 + nmax * 0.00625e12
```
jktjkt:
```python
center = 193.1e12
freq_min = center + nmin * grid
freq_max = center + nmax * grid
```

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+    print('initial spectrum')
+    print(nvalue_to_frequency(oms_list[10].spectrum_bitmap.freq_index_min) * 1e-12,
+          nvalue_to_frequency(oms_list[10].spectrum_bitmap.freq_index_max) * 1e-12)
```
jktjkt: This looks like leftovers from debugging? If you want to check them, let's make this explicit via asserts, it's a unit test after all.

Also at other places below.

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+    # checks initial values consistancy
+    ind_max = len(oms_list[10].spectrum_bitmap.bitmap) - 1
```
jktjkt: I don't know what `10` means. It's used at a lot of places below, so let's give it a name so that a reader can keep up, e.g.:

```python
last_fiber = oms_list[10]
# ...
assert nvalmin == last_fiber.spectrum_bitmap.freq_index_min
```

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+@pytest.mark.parametrize('nval2', [8, 12])
+def test_assign_and_sum(nval1, nval2, setup):
+    """ checks that bitmap sum gives correct result
+    """
+    network, oms_list = setup
+    grid = 0.00625e12
+    guardband = grid
+    mval = 4 # slot in 12.5GHz
+    freq_min = 193.1e12
+    freq_max = 193.1e12 + 24 * 0.00625e12
```
jktjkt: some of these constants (also see comment for the previous test) look like something that can be move to the scope of the whole file

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+        for elem in range1:
+            print(f'value should be zero at index {elem}')
+            assert test2[elem] == 0
+
+        for elem in range2:
+            print(f'value should be zero at index {elem}')
+            assert test2[elem] == 0
+
+def test_values(setup):
+    """ checks that oms.assign_spectrum(13,7) is (193137500000000.0, 193225000000000.0)
```
jktjkt: This looks like an excellent use case for a docstring. Here's [an example how it's used in `gnpy.core.utils`](https://github.com/Telecominfraproject/oopt-gnpy/blob/develop/gnpy/core/utils.py#L187); these comments are auto-extracted by the test framework and run as a part of the test suite. No need to write the test in `tests/`, just add two lines to the method's docstring.

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+def test_exception(nval, mval, setup):
+    """ test n or m not applicable values
+    """
+    network, oms_list = setup
+    try:
+        oms_list[5].assign_spectrum(nval, mval)
+        print(f'n, m values should raise an error {nval}, {mval}')
+        test = False
+    except SpectrumError:
+        test = True
```
jktjkt: Pytest has a way of saying that [this code should always throw](https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions), that should be used here.

This test is, however, a bit more complicated because it is somehow expected to *not* fail when `m + n = 1`. That's non-obvious to me, and I think it would be better to split this into two separate tests, one for correct values which should not throw, and one for expected failures.

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+    """ test n or m not applicable values
+    """
+    network, oms_list = setup
+    test = oms_list[5].assign_spectrum(nval, mval)
+    print(f'n, m values should raise an error {nval}, {mval}')
+    expected = False
+    if (nval + mval) == 1:
+        expected = True
+    print(nval, mval)
+    assert test is expected
```
jktjkt: The `is` operator tests for equivalent object IDs, not for equivalent values. I think that it indeed works with bools, but the test should still use `==` in that context, IMHO.

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+    assert test or (nval + mval) == 1
+
+@pytest.mark.parametrize('nval', [0, -300, 500])
+@pytest.mark.parametrize('mval', [1, 600])
+def test_wrong_values(nval, mval, setup):
+    """ test n or m not applicable values
+    """
+    network, oms_list = setup
+    test = oms_list[5].assign_spectrum(nval, mval)
+    print(f'n, m values should raise an error {nval}, {mval}')
```
jktjkt: I don't see any exception raised here. See above for a suggestion on how to write these.

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+        expected = True
+    print(nval, mval)
+    assert test is expected
+
+def test_bitmap_assignment(setup):
+    """ test that a bitmap can be assigned
+    """
+    network, oms_list = setup
+
+    oms_list[5].assign_spectrum(13, 7)
```
jktjkt: magic `5`, let's give it a descriptive name, even if it is just a (rather non-descriptive) `element`

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+
+    btmp = deepcopy(oms_list[5].spectrum_bitmap.bitmap)
+    freq_min = 191300000000000.0
+    freq_max = 196100000000000.0
```
jktjkt: use a file-global constant (also everywhere else, and for grid spacing and guardbands)

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+    # try a wrong asignment that should not pass
+    btmp = btmp[1:-1]
+    test = False
+    try:
+        spectrum_btmp = Bitmap(freq_min, freq_max, grid=0.00625e12, guardband=0.15e12, bitmap=btmp)
+    except SpectrumError:
+        test = True
+
+    print('bitmap direct assignment should create an error if length is not consistant with' +\
+          'provided values')
```
jktjkt: I think that it's better to use a longer line rather than split an error message into two just for two extra words. Let's address this once we move to Gerrit (with a testsuite's built-in PEP8 checking).

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+    for nval in range(100):
+        (center_n, startn, stopn), path_oms = spectrum_selection(pths[0], oms_list, 4)
+        pth_assign_spectrum(pths, req, oms_list, [find_reversed_path(pths[0])])
+        print(f'testing on following oms {path_oms}')
+        # check that only 96 channels are feasible
+        if nval >= 96:
+            print(center_n, startn, stopn)
+            print('only 96 channels of 4 slots pass in this grid')
+            assert center_n is None and startn is None and stopn is None
+        if nval < 96:
```
jktjkt: `else`

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+            print(center_n, startn, stopn)
+            print('at least 96 channels of 4 slots should pass in this grid')
+            assert center_n is not None and startn is not None and stopn is not None
+        # reset req[0]: since pth_assign_spectrum adds N and M attribute to the class,
+        # removing N and M simulates a brand new request at each loop.
+        delattr(req[0],'N')
+        delattr(req[0],'M')
```
jktjkt: I think that `@pytest.parametrize` is better -- the currently chosen approach requires an extra explanation

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+    req = [rqs[2]]
+    pths = compute_path_dsjctn(network, equipment, req, [])
+    (center_n, startn, stopn), path_oms = spectrum_selection(pths[0], oms_list, 4, 478)
+    print(oms_list[0].spectrum_bitmap.freq_index_max)
+    print(oms_list[0])
+    print(center_n, startn, stopn)
+    print('spectrum selection error: should be None')
+    assert center_n is  None and startn is None and stopn is None
```
jktjkt: nit: extra space

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+from json import loads
+from math import ceil
+import pytest
+from gnpy.core.equipment import load_equipment, automatic_nch
+from gnpy.core.network import load_network, build_network
+from gnpy.core.utils import lin2db
+from gnpy.core.elements import Roadm, Transceiver
+from gnpy.core.spectrum_assignment import (build_oms_list, align_grids, nvalue_to_frequency,
+                                           bitmap_sum, m_to_freq, slots_to_m, frequency_to_n,
+                                           Bitmap, spectrum_selection, pth_assign_spectrum)
```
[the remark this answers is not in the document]
EstherLerouzic: I have tried, but then I need to put a prefix on each function call which I find a bit cumbersome:

import gnpy.core.spectrum_assignment as sp
sp.nvalue_to_frequency(....

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+
+"""
+@author: esther.lerouzic
+
+"""
+
+from pathlib import Path
+from copy import deepcopy
+from json import loads
+from math import ceil
```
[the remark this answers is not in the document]
EstherLerouzic: OK, done

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+    return equipment
+
+@pytest.fixture()
+def setup(eqpt):
+    """ common setup for tests: builds network, equipment and oms only once
+    """
+    equipment = eqpt
+    # fix band to be independant of changes in json file
+    equipment['SI']['default'].f_min = 191300000000000.0
+    equipment['SI']['default'].f_max = 196100000000000.0
```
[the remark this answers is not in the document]
EstherLerouzic: OK, agreed!

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+@pytest.fixture()
+def eqpt():
+    """ common setup for tests: builds network, equipment and oms only once
+    """
+    equipment = load_equipment(EQPT_FILENAME)
+    return equipment
+
+@pytest.fixture()
+def setup(eqpt):
+    """ common setup for tests: builds network, equipment and oms only once
```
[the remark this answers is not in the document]
EstherLerouzic: OK, done

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+    """ tests that the oms is between two roadms, that there is no roadm or transceivers in the oms
+        except end points, checks that the id of oms is present in the element and that the element
+        oms id is consistant
+    """
+    network, oms_list = setup
+    for oms in oms_list:
+        assert isinstance(oms.el_list[0], Roadm) and isinstance(oms.el_list[-1], Roadm)
+        for i, elem in enumerate(oms.el_list[1:-2]):
+            assert not isinstance(elem, Roadm)
+            assert elem  in network.nodes()
```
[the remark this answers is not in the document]
EstherLerouzic: OK done

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+            assert elem.oms.oms_id == oms.oms_id
+            print(f'expected {elem.uid}, obtained {oms.el_id_list[i+1]}')
```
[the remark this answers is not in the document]
EstherLerouzic: OK ! thanks!

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+@pytest.mark.parametrize('nmax', [480, 320, 450])
+def test_aligned(nmin, nmax, setup):
+    """ checks that the oms grid is correctly aligned. Note that bitmap index uses guardband
+        on both ends so that if nmin, nmax = -200, +200, min/max navalue in bitmap are
+        -224, +223, which makes 223 -(-224) +1 frequencies
+    """
+    network, oms_list = setup
+    # f_min = 193.1e12 - 280 * 0.00625e12
+    # f_max = 193.1e12 + 320 * 0.00625e12
+    # f_min = 1931.1 f_max = 195.1
```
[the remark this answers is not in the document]
EstherLerouzic: just a reminder for me. I removed this

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+    """
+    equipment = eqpt
+    # fix band to be independant of changes in json file
+    equipment['SI']['default'].f_min = 191300000000000.0
+    equipment['SI']['default'].f_max = 196100000000000.0
+    network = load_network(NETWORK_FILENAME, equipment)
+    p_db = equipment['SI']['default'].power_dbm
+    p_total_db = p_db + lin2db(automatic_nch(equipment['SI']['default'].f_min,\
+        equipment['SI']['default'].f_max, equipment['SI']['default'].spacing))
+    build_network(network, equipment, p_db, p_total_db)
```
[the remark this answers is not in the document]
EstherLerouzic: OK, Done

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+
+def test_oms(setup):
+    """ tests that the oms is between two roadms, that there is no roadm or transceivers in the oms
+        except end points, checks that the id of oms is present in the element and that the element
+        oms id is consistant
+    """
+    network, oms_list = setup
+    for oms in oms_list:
+        assert isinstance(oms.el_list[0], Roadm) and isinstance(oms.el_list[-1], Roadm)
+        for i, elem in enumerate(oms.el_list[1:-2]):
```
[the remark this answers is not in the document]
EstherLerouzic: OK ! thanks! That was a mistake !

tests/test_spectrum_assignment.py
```diff
@@ -0,0 +1,378 @@
...
+        on both ends so that if nmin, nmax = -200, +200, min/max navalue in bitmap are
+        -224, +223, which makes 223 -(-224) +1 frequencies
+    """
+    network, oms_list = setup
+    # f_min = 193.1e12 - 280 * 0.00625e12
+    # f_max = 193.1e12 + 320 * 0.00625e12
+    # f_min = 1931.1 f_max = 195.1
+    grid = 0.00625e12
+    guardband = 0.15e12
+    nguard = 24
```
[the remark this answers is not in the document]
EstherLerouzic: OK Done

tests/test_spectrum_assignment.py
EstherLerouzic: OK, Done

tests/test_spectrum_assignment.py
EstherLerouzic: OK ! thanks!

tests/test_spectrum_assignment.py
EstherLerouzic: OK Done

tests/test_spectrum_assignment.py
EstherLerouzic: Thanks for the suggestion I added them + some more

tests/test_spectrum_assignment.py
EstherLerouzic: Thanks! good to know! also enabled me to add a test on negative values for m (not allowed) :)

tests/test_spectrum_assignment.py
EstherLerouzic: Thanks, I corrected!

tests/test_spectrum_assignment.py
EstherLerouzic: Magic 5 replace with several other randomly picked magic indexes, thanks for the suggestion !

tests/test_spectrum_assignment.py
EstherLerouzic: Done!

tests/test_spectrum_assignment.py
EstherLerouzic: OK!

tests/test_spectrum_assignment.py
EstherLerouzic: OK!

tests/test_spectrum_assignment.py
EstherLerouzic: That one was a bit more difficult to handle with parametrize, because the networks need to be loaded at each loop of nval, and a parametrize deletes previous loads. I used a deepcopy of the request instead to avoid delattr

tests/test_spectrum_assignment.py
EstherLerouzic: ok

PR #304, opened 2019-09-23
Give more useful comment for user to correct topology

EstherLerouzic:
when a node name used in link does not correspond to listed
node names, Bad link msg was thrown without info on which link
causes the problem.
This change gives additional info on link and catch the user error
more cleanly.

### DIFF
--- a/gnpy/core/convert.py
+++ b/gnpy/core/convert.py
@@ -32,6 +32,7 @@ from json import dumps
 from pathlib import Path
 from difflib import get_close_matches
 from gnpy.core.utils import silent_remove
+from gnpy.core.exceptions import NetworkTopologyError
 import time
 
 all_rows = lambda sh, start=0: (sh.row(x) for x in range(start, sh.nrows))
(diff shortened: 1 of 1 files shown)


gnpy/core/convert.py
```diff
@@ -511,7 +512,11 @@ def parse_excel(input_filename):
         raise ValueError(f'Duplicate city: {all_cities}')
     if any(ln.from_city not in all_cities or
            ln.to_city   not in all_cities for ln in links):
-        raise ValueError(f'Bad link.')
+        bad_links = []
+        for lnk in links:
+            if lnk.from_city not in all_cities or lnk.to_city not in all_cities:
+                bad_links.append([lnk.from_city, lnk.to_city])
```
EstherLerouzic: Is this better like this ?

jktjkt: The code checks all the links twice (once in the `any(...)`, then once more in the explicit loop). The benefit of this is that `bad_links` can be scoped. I don't really care that much, but if I was writing this, I would probably do it like this:

```py
bad_links = []
for link in links:
  if ...:
    bad_links.append((link.from_city, link.to_city))
if not bad_links.empty():
  raise NetworkTopologyError(...)
```

Feel free to approve and merge as-is if you prefer the existing patch.

EstherLerouzic: understood , thanks

gnpy/core/convert.py
```diff
@@ -509,9 +510,12 @@ def parse_excel(input_filename):
...
     if len(all_cities) != len(nodes):
         raise ValueError(f'Duplicate city: {all_cities}')
-    if any(ln.from_city not in all_cities or
-           ln.to_city   not in all_cities for ln in links):
-        raise ValueError(f'Bad link.')
+    bad_links = []
+    for lnk in links:
+        if lnk.from_city not in all_cities or lnk.to_city not in all_cities:
+            bad_links.append([lnk.from_city, lnk.to_city])
+    if bad_links:
```
jktjkt: nice touch.

PR #305, opened 2019-09-23
Reading pout target column in excel Nodes sheet

EstherLerouzic:
this reads the value per ROADM node and insert it in
target_pch_out_db of the roadm node instance.
value is ignored if node type is not ROADM.


EstherLerouzic:
close this one and will propose a solution to also account for per degree target power

PR #306, opened 2019-09-23
docs: show a DOI

jktjkt:
As per @aleFerrari 's request, this adds a DOI and makes this software citeable. DOIs are assigned automatically via Zenodo in a process that's integrated with GitHub.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -7,7 +7,7 @@
 `gnpy`: mesh optical network route planning and optimization library
 ====================================================================
 
-|docs| |build|
+|docs| |build| |doi|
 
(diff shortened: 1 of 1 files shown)


jktjkt:
For some reason the DOI link does not appear to work, despite what Zenodo says. Let's give this a few days to see if it actually works.

ojnas:
The link works for me.

jktjkt:
> The link works for me.

Thanks -- it was apparently just a matter of time for the record to propagate from Zenodo infrastructure to doi.org.

PR #308, opened 2019-09-27
Rework OSNR printing

jktjkt:
Esther and Jonas pointed out that it is much more common to work with OSNR measurements per 0.1nm rather than per signal bandwidth. This is in line with what OSAs usually report and what transponder datasheets specify.

cc #307

### DIFF
--- a/examples/transmission_main_example.py
+++ b/examples/transmission_main_example.py
@@ -161,7 +161,7 @@ def main(network, equipment, source, destination, sim_params, req=None):
             print(f'\nTransmission result for input power = {lin2db(req.power*1e3):.2f} dBm:')
         else:
             print(f'\nTransmission results:')
-        print(f'  Final SNR total (signal bw): {ansi_escapes.cyan}{mean(destination.snr):.02f} dB{ansi_escapes.reset}')
+        print(f'  Final SNR total (0.1 nm): {ansi_escapes.cyan}{mean(destination.snr_01nm):.02f} dB{ansi_escapes.reset}')
 
         #print(f'\n !!!!!!!!!!!!!!!!!     TEST POINT         !!!!!!!!!!!!!!!!!!!!!')
         #print(f'carriers ase output of {path[1]} =\n {list(path[1].carriers("out", "nli"))}')
--- a/examples/transmission_main_example.py
+++ b/examples/transmission_main_example.py
@@ -157,13 +157,13 @@ def main(network, equipment, source, destination, sim_params, req=None):
         if len(power_range) == 1:
             for elem in path:
                 print(elem)
+            if power_mode:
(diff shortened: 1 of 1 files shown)


jktjkt:
@EstherLerouzic , does this address your concerns from #307?

EstherLerouzic:
Yes it addresses my concern, but the cyan color is attracting the eye on power instead of final SNR, so I propose to color the final SNR in o.1 nm in the same way.
Moreover I tried the program on a colleague and it was not clear to him that the final value really integrates tx OSNR and add drop osnr, since the first element shows +inf values. So I propose to add a line to better explain this on the final result. I have posted my proposal here (https://github.com/Orange-OpenSource/gnpy/tree/osnr-bw-printing)

ojnas:
> since the first element shows +inf values

I think this is kind of a separate issue and is something I have thought about before. It is confusing to show +inf when the user has specified a tx_osnr.

> So I propose to add a line to better explain this on the final result.

I disagree with this proposal. It should not have to be explained that the **final** SNR includes all contributions. I would rather change the printing of the first transceiver to include tx_osnr (not add_drop_osnr since this is a property of the ROADM, not the transceiver), see my proposal in PR #311.

EstherLerouzic:
> I disagree with this proposal. It should not have to be explained that the **final** SNR includes all contributions. I would rather change the printing of the first transceiver to include tx_osnr (not add_drop_osnr since this is a property of the ROADM, not the transceiver), see my proposal in PR #311.

I understand your point and I agree that is is a better solution for users to read the results. I think however that we should add comments in the code because it may be confusing for a newcomer to see that the update is performed twice... Something like:
'tx_osnr and add_drop_osnr are added at the end of the propagation, but the tx_osnr contribution is added on the emitter element after propagation for cleaner printings.'

jktjkt:
> Yes it addresses my concern, but the cyan color is attracting the eye on power instead of final SNR, so I propose to color the final SNR in o.1 nm in the same way.

I agree with the motivation, but I'm afraid that [post-processing the output of `Transceiver.__str__`](https://github.com/Orange-OpenSource/gnpy/commit/e50cf1f775026b136469e19be96565340a36248d#diff-25465525343a2233183dc73986d8af2fR168-R171) is too fragile, I really do not want to merge that. I propose to leave this as-is for now.

> Moreover I tried the program on a colleague and it was not clear to him that the final value really integrates tx OSNR and add drop osnr, since the first element shows +inf values. So I propose to add a line to better explain this on the final result. I have posted my proposal here (https://github.com/Orange-OpenSource/gnpy/tree/osnr-bw-printing)

Let's continue this discussion at #311. It seems to me that the real issue is that the OSNR is not reflected in the signal from the very beginning.

PR #309, opened 2019-10-02
add the option "rest" to activate the api-rest

EstherLerouzic:
On behalf of @ahmed-triki 
This is an implementation of an API rest for GNPy. The http server is based on flask, so the installation of this package is mandatory. The input/output of the api-rest are modeled by a yang data model.
In addition to the Rest mode, the new code keeps the possibility to use the current CLI (Command Line Interface) mode.
To activate the Rest mode, the user has to launch the following command :
python path_requests_run.py --rest
The API-rest client has to take into account the following connection parameters:

    URL: http://<ip_address>:5000/gnpy/api/v1.0/files
    authentication: username: gnpy/ password: gnpy
    Rest-Method: POST / The body of the request is composed of the topology and the service data in json format.
    
    Todo: add explanation of the body and the response models of the API-REST.

-------
- add the function launch_cli to launch the "cli" mode
- add the the class Gnpy_API to launch the "api" mode
- modify the main to enable the launch of Gnpy with two
modes "rest" and "cli"

Signed-off-by: ahmed <ahmed.triki@orange.com>

### DIFF
--- a/examples/path_requests_run.py
+++ b/examples/path_requests_run.py
@@ -23,7 +23,7 @@ from networkx import (draw_networkx_nodes, draw_networkx_edges,
 from numpy import mean
 from gnpy.core.service_sheet import convert_service_sheet, Request_element, Element
 from gnpy.core.utils import load_json
-from gnpy.core.network import load_network, build_network, save_network
+from gnpy.core.network import load_network, network_from_json, build_network, save_network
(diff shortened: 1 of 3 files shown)


EstherLerouzic:
To be correced and rebased after other PRs ( eg#303)

PR #311, opened 2019-10-04
Include tx_osnr when printing first transceiver

ojnas:
Currently the output of transmission_main_example.py shows the first transceiver having ONSR = +inf even when a tx_osnr has been specified, which is confusing.
This patch proposes to update the SNR of the first transceiver with tx_osnr before printing.

### DIFF
--- a/gnpy/core/request.py
+++ b/gnpy/core/request.py
@@ -399,6 +399,7 @@ def propagate(path, req, equipment):
         req.power, req.spacing)
     for el in path:
         si = el(si)
+    path[0].update_snr(req.tx_osnr)
     path[-1].update_snr(req.tx_osnr, equipment['Roadm']['default'].add_drop_osnr)
     return path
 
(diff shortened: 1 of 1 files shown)


jktjkt:
Is there a reason why `Transceiver.__call__` does not simply introduce an appropriate amount of ASE noise to each channel of the `SpectrumInformation`? That way, there would be no need to mess with incremental SNR additions.

To me, it's rather counter intuitive that it's OK to perform the whole simulation/propagation, and just before we're about to print the result, we magically adjust the SNR at the start. Surely the SNR "present from the beginning" must have had some effect along the path as well?

ojnas:
@jktjkt, good question, I agree that it's counter-intuitive.

The ASE noise can't be introduced in `Transceiver.__call__`, since then it would be introduced at both the transmit and receive ends.

My original proposal (PR #177), which was merged but later reverted, was to introduce the transmitter ASE noise in  `create_input_spectral_information`. It turned out that this increased the computation time of the `propagate_and_optimize_mode` function, see discussion in PR #186 and issue #185. I then proposed an optimization in PR #197 but that was rejected in favor of instead adding the transmitter ASE noise after propagation, see PR #199.

jktjkt:

Issue #132, opened 2018-10-23
Replace `collections.namedtuple` with `dataclasses` or `attrs`


dutc:
Referencing #100, we need to improve our `Elements` base class to better handle optional fields while still validating required fields.

There are two major options we can choose for this:
- `dataclasses` (built into Python >3.7, available via `pip install dataclasses` prior to that) https://www.python.org/dev/peps/pep-0557/
- `attrs` (user-supported library) https://github.com/python-attrs/attrs

We want to preserve the immutable behavior of network elements for ease-of-reasoning and optimisation purposes. We want to improve the user interface when creating new fields and improve the robustness of our code in validating user input on required fields.

ojnas:
I have a couple of questions related to the "immutable behavior of network elements":

1. Why is not the Edfa element class using an immutable namedtuple for `params` like all the other element classes? This looks a bit inconsistent.

2. What exactly is the importance of this immutable behavior? Currently, for other elements than Edfa, the immutable `params` is anyway not used after initialization, where it is just used for assigning values to other (mutable) variables. For example in `class Roadm` you have `self.loss = self.params.loss`. Then this `roadm.loss` variable is updated by `build_network` in `network.py`. Same for Fiber class variables `con_in` and `con_out` for example.

jeanluc-auge:
I have check the data class: I like it a lot.
Only drawback it needs python 3.7, which is a bit of a pain to update.

PR #319, opened 2019-12-03
Polito restructuring parameters

AndreaDAmico:


ojnas:
What is the motivation for replacing the element params initialization based on namedtuple with something else? I think the original motivation for using namedtuple was that immutable behavior of network elements was desired but see also issue #132.

gnpy/core/node.py
```diff
@@ -26,7 +26,7 @@ def __new__(cls, latitude=0, longitude=0, city=None, region=None):
         return super().__new__(cls, latitude, longitude, city, region)
 
 class Node:
-    def __init__(self, uid, name=None, params=None, metadata=None, operational=None):
+    def __init__(self, uid, name=None, params=None, metadata=None, operational=None, type_variety=None):
```
ojnas: What is the motivation for introducing `type_variety` here?

gnpy/core/elements.py
```diff
@@ -335,25 +302,28 @@ def carriers(self, loc, attr):
...
-        Dispersion is entered in ps/nm/km.
-        Disperion can be a numpy array or a single value.
+    def alpha(self, frequencies):
+        """ It returns the values of the series expansion of attenuation coefficient alpha(f) for all f in frequencies
+
+        :param frequencies: frequencies of series expansion [Hz]
+        :return: alpha: power attenuation coefficient for f in frequencies [Neper/m]
+        """
+        if type(self.params.loss_coef) == dict:
+            alpha_interp = interp1d(self.params.lin_loss_exp, self.params.f_loss_ref)
```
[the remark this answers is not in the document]
AndreaDAmico: This has been addressed - numpy.interp is now used instead of scipy.interpolate.interp1d. This method was original in fiber_params in science_utils.py and had been moved without editing.

gnpy/core/elements.py
```diff
@@ -224,37 +226,15 @@ def __call__(self, spectral_info):
         pref = self.update_pref(spectral_info.pref)
         return spectral_info._replace(carriers=carriers, pref=pref)
 
-FiberParams = namedtuple('FiberParams', 'type_variety length loss_coef length_units \
-                                         att_in con_in con_out dispersion gamma')
 
 class Fiber(Node):
     def __init__(self, *args, params=None, **kwargs):
-        if params is None:
+        if not params:
```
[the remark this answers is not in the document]
AndreaDAmico: None, empty lists, dictionaries and tuples are interpreted as False under boolean conditions. In this case, the if condition is used to handle the exception within the FiberParams object instead of raising an Exception when the ** unpacker is called.

gnpy/core/elements.py
```diff
@@ -224,37 +226,15 @@ def __call__(self, spectral_info):
...
-        self.att_in = self.params.att_in
-        self.con_in = self.params.con_in
-        self.con_out = self.params.con_out
-        self.dispersion = self.params.dispersion  # s/m/m
-        self.gamma = self.params.gamma # 1/W/m
-        self.pch_out_db = None
         self.carriers_in = None
         self.carriers_out = None
-        # TODO|jla: discuss factor 2 in the linear lineic attenuation
+        self.nli_solver = NliSolver(self)
```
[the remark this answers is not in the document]
AndreaDAmico: This has been addressed in the recent commit e33cfa3. To clarify, this attribute is currently only used within science_utils.propagate_raman_fiber.

gnpy/core/elements.py
```diff
@@ -413,46 +384,16 @@ def __call__(self, spectral_info):
...
-    @sim_params.setter
-    def sim_params(self, sim_params=None):
-        self._sim_params = sim_params
+        super().__init__(*args, params=params, **kwargs)
+        if 'raman_pumps' in self.operational:
+            self.raman_pumps = tuple(PumpParams(p['power'], p['frequency'], p['propagation_direction'])
+                  for p in self.operational['raman_pumps'])
+        else:
+            self.raman_pumps = None
+        self.raman_solver = RamanSolver(self)
```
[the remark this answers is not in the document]
AndreaDAmico: This has been addressed in the recent commit 579d7fb.

gnpy/core/node.py
AndreaDAmico: In the current version type_variety is duplicated: it is present as an attribute of Fiber and of Fiber.params. We want to get rid of this redundancy; we believe that as type_variety is a description rather than a physical parameter of the fiber, the right place should be outside of Fiber.params. Moreover, we think that the right way to include this attribute in any element is to define it in the Node constructor.

AndreaDAmico:
> What is the motivation for replacing the element params initialization based on namedtuple with something else? I think the original motivation for using namedtuple was that immutable behavior of network elements was desired but see also issue #132.

The primary motivation for restructuring the code is that we would like to have a single object which contains all of the parameters required for each Fiber instance. Some parameters are set by the namedtuple definitions within the modules, whereas others are evaluated and set during runtime. We have defined a class which contains all of these parameters. 
Concerning mutability, we declare all attributes that we wish to keep fixed as properties without a setter, reproducing the namedtuple immutability behaviour. With regards to the runtime modified attributes, we define a setter in order to allow their modification. This feature may cause some mutability/immutability issues if the assignment of a label is not made with the constructor of fiber.params (something which is not present within the current code).
In the end, we can recover the namedtuple immutability for all attributes of fiber.params by replacing the setter with a "replace" method that returns a new instance with the modified attributes.
@jktjkt @EstherLerouzic : If you agree with this solution we can implement it within this pull request.

gnpy/core/elements.py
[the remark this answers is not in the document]
ojnas: If you only use it in the `RamanFiber` class, why do you instantiate `NliSolver` in the `Fiber` class?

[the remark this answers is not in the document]
AndreaDAmico: We do this as we intend to restructure the propagation of the fiber (currently fully defined within element.fiber) to instead use the NLISolver (to avoid again a duplication of the code).

ojnas:
> Concerning mutability, we declare all attributes that we wish to keep fixed as properties without a setter, reproducing the namedtuple immutability behaviour. With regards to the runtime modified attributes, we define a setter in order to allow their modification.

Ok, even though all attributes can of course still be modified by accessing the underscore prefixed attribute names directly, I believe this follows Python conventions and is an improvement over the existing code that anyway treats immutable attributes as mutable (e.g. by copying them to new attributes as is done in the Fiber class).

tests/test_parameters.py
```diff
@@ -0,0 +1,37 @@
...
+
+import json, pytest
+
+from gnpy.core.parameters import SimParams
+from gnpy.core.science_utils import Simulation
+from gnpy.core.elements import Fiber
+
+
+def test_sim_parameters():
+    f = open('tests/data/test_sim_params.json')
```
EstherLerouzic: could be better to use relative path here 

```
from pathlib import Path
TEST_DIR = Path(__file__).parent

with open(TEST_DIR / 'tests' / 'data' / 'test_sim_params.json'):
   ...
```
>

examples/sim_params.json
```diff
@@ -9,6 +8,7 @@
   	"nli_method_name": "ggn_spectrally_separated",
   	"wdm_grid_size": 50e9,
   	"dispersion_tolerance": 1,
-  	"phase_shift_tollerance": 0.1
+  	"phase_shift_tolerance": 0.1,
+    "computed_channels": [1, 18, 37, 56, 75]
```
EstherLerouzic: not the correct indent ?

gnpy/core/elements.py
```diff
@@ -263,62 +242,49 @@ def to_json(self):
...
                 'params'        : {
                 #have to specify each because namedtupple cannot be updated :(
-                    'type_variety'  : self.type_variety,
-                    'length'        : self.length/UNITS[self.params.length_units],
-                    'loss_coef'     : self.loss_coef*1e3,
-                    'length_units'  : self.params.length_units,
-                    'att_in'        : self.att_in,
-                    'con_in'        : self.con_in,
-                    'con_out'       : self.con_out
+                    'length'        : round(self.params.length * 1e-3, 6),
```
EstherLerouzic: This removes
`self.length/UNITS[self.params.length_units]`
 Does this means that lengths are in km and unit is no more a relevant attribute in json ? then README should be corrected

AndreaDAmico: The length unit is still useful in the configuration .json because the user can express the length in any unit he/she prefers using this parameter. No matter what their choice is, the length of the fiber instance is converted into meters as a fundamental unit, meaning that the length_unit fiber instance parameter always end up as 'm'.
In the quoted piece of code, a json representative of the fiber is created. We think that expressing length in km is a reasonable choice to support legibility.

EstherLerouzic: OK, I understand, however if I use an input json network file with units in m, then transmission_main does not work anymore:
OverflowError: (34, 'Numerical result out of range')
 I think that the initialization of Fiber in
`network_from_json(json_data, equipment)`
leaves the possibility to have other units, so that 
`self.params.length` can have different units, and maybe this gives issues in the initialization of params ?

AndreaDAmico: The point here is that length_unit should concern all quantities in fiber parameters that include a unit of length. This is also the case for loss_coef; if 'm' is chosen as length_unit, loss_coeff should be expressed in dB/m in the configuration .json.

ojnas: This happens because unit conversion is applied also to `loss_coef` while in the existing code it is assumed that `loss_coef` is always specified in dB/km.

ojnas: > The point here is that length_unit should concern all quantities in fiber parameters that include a unit of length.

But you don't apply conversion to dispersion and gamma which also include a unit of length.

AndreaDAmico: > > The point here is that length_unit should concern all quantities in fiber parameters that include a unit of length.
> 
> But you don't apply conversion to dispersion and gamma which also include a unit of length.

This is true. Nevertheless, since both dispersion and gamma are already expressed in fundamental units, we thought that this would be always the case as a requirement.

EstherLerouzic: Right now, the user has the flexibility to use units in the topo file for the length only, independently from the units of the lineic loss which is assumed to be always in dB/km. I would prefer to keep this as it is in this PR (changing the `/ self._length_units_factor`  into a `/ UNITS['km']` in parameters.py, lines 169 to 175). The question on meter or km is trivial, but some user may have prepared all their files in meters and I would prefer to avoid a refactor change that also changes user experience. Would it be possible to postpone this discussion/(unit uniformisation) when we will discuss yang ?

gnpy/core/network.py
```diff
@@ -534,8 +531,7 @@ def build_network(network, equipment, pref_ch_db, pref_total_db):
     for fiber in fibers:
         split_fiber(network, fiber, bounds, target_length, equipment)
 
-    amplified_nodes = [n for n in network.nodes()
-                        if isinstance(n, Fiber) or isinstance(n, Roadm)]
+    amplified_nodes = [n for n in network.nodes() if isinstance(n, Fiber) or isinstance(n, Roadm)]
```
EstherLerouzic: This list is to identify nodes that will need to have a amplifier (egress) in case the user did not put one (autodesign): shouldn't there be a  `or isInstance(n, RamanFiber)` here ?

examples/name_match_dictionary.json
```diff
@@ -0,0 +1 @@
+{}
```
EstherLerouzic: I think this file should not be part of the commits

tests/name_match_dictionary.json
```diff
@@ -0,0 +1 @@
+{}
```
EstherLerouzic: I think this file should not be part of commits

gnpy/core/network.py
AndreaDAmico: Since RamanFiber inherit from Fiber, a RamanFiber instance would be also a Fiber instance

gnpy/core/elements.py
AndreaDAmico: The choice of 'm' or 'km' involves some subtlety - in the current version's configuration json the lineic loss is always assumed to be in dB/km, however, we want the fiber parameter loss_coef to be expressed in fundamental units (dB/m). This means that it is not possible to avoid a duplication of this parameter in this case; we thought that using the length_units would solve this issue. In any case, in the last commit we restore this duality of loss_coef units (dB/km in the configuration files, dB/m in the fiber instance) in order to guarantee no changes for the user, but this introduces an ambiguity between the instance's value and the class when called as a dictionary (using asdict).

jktjkt:
@EstherLerouzic , during the Tuesday call we discussed whether it is OK to take the sequence of commits "as is", with additional fixes going in as follow-up commits. I suggested to the Polito team to split their changes into self-contained commits -- going from a working state to another still working state, changing just one small thing at a time, etc. I pointed to a non-linear history with merges, duplicate changes ([`.idea` take one](https://github.com/Telecominfraproject/oopt-gnpy/pull/319/commits/fb53b4f348d06740bd4af49f8d6ceb7d3f8a8df9), [`.idea` take two](https://github.com/Telecominfraproject/oopt-gnpy/pull/319/commits/fb53b4f348d06740bd4af49f8d6ceb7d3f8a8df9)), commits which perform unrelated changes in one go, etc.

(One example which I did not mention during the call were [fix-ups for regressions introduced in this refactoring](https://github.com/Telecominfraproject/oopt-gnpy/pull/319/commits/2d4428cb70a1d04b81a0b5549444577b3acb8ed0). That one effectively breaks `git bisect`, and I'm therefore against taking this PR without rewriting the git history.)

Vittorio expressed their preference for taking this PR as-is. I wanted to hear your opinion about this one. Should we insist on this refactoring "done right as Jan sees it" (and therefore reworking this PR into smaller chunks, one thing at a time), or are you as a co-maintainer OK with a git history with fix-ups?

EstherLerouzic:
> 
> 
> @EstherLerouzic , during the Tuesday call we discussed whether it is OK to take the sequence of commits "as is", with additional fixes going in as follow-up commits. I suggested to the Polito team to split their changes into self-contained commits -- going from a working state to another still working state, changing just one small thing at a time, etc. I pointed to a non-linear history with merges, duplicate changes ([`.idea` take one](https://github.com/Telecominfraproject/oopt-gnpy/pull/319/commits/fb53b4f348d06740bd4af49f8d6ceb7d3f8a8df9), [`.idea` take two](https://github.com/Telecominfraproject/oopt-gnpy/pull/319/commits/fb53b4f348d06740bd4af49f8d6ceb7d3f8a8df9)), commits which perform unrelated changes in one go, etc.
> 
> (One example which I did not mention during the call were [fix-ups for regressions introduced in this refactoring](https://github.com/Telecominfraproject/oopt-gnpy/pull/319/commits/2d4428cb70a1d04b81a0b5549444577b3acb8ed0). That one effectively breaks `git bisect`, and I'm therefore against taking this PR without rewriting the git history.)
> 
> Vittorio expressed their preference for taking this PR as-is. I wanted to hear your opinion about this one. Should we insist on this refactoring "done right as Jan sees it" (and therefore reworking this PR into smaller chunks, one thing at a time), or are you as a co-maintainer OK with a git history with fix-ups?

@jktjkt,
I think that the git history is a bit messy and would benefit from cleaning if one wants to read commit per commit or run git bisect. But I do not think that having all these commits is needed, since this is a refactor: it is hard to cut it into pieces (maybe except unrelated pieces where the automatic linter added changes such as line cuts or spaces). 

- To speed delivery, I think that one possibility would be to squash all commits and have this PR as a single commit. In this way git bisect would not break. 

- If the delivery is not urgent, I would also recommend to clean git history. For example, separate the unrelated changes (.gitignore, path_requests_run and all linter corrections), the code change . the test changes.

In general, the recommendation I have been given by my colleagues is to keep as much as possible the following rule:

   - one PR = one commit

   - and if there are more than one commit, each of them must be self consistant, ie no need to read them all to understand, and each commit in a state where the program can run ("going from a working state to another still working state"). 

Speed of delivery has been a key factor for GNPy last year, and my preference would be not to delay too much. I am not sure that having a cleaner history will help catching bugs if any. Adding tests would help better.

EstherLerouzic:
I have found another issue:
before refactor if I try to launch an example file  with a raman fiber without sim_param option:
```
(gnpy) user@esther-b:~/workspace/TIP/gnpy/examples$ python transmission_main_example.py ramantest.json 
There are 76 channels propagating
Power mode is set to True
=> it can be modified in eqpt_config.json - Span
Invocation error: RamanFiber requires passing simulation params via --sim-params
```
after refactor:

```
(gnpy) user@esther-b:~/workspace/TIP/gnpy/examples$ python transmission_main_example.py ramantest.json 
Traceback (most recent call last):
  File "transmission_main_example.py", line 222, in <module>
    network = load_network(args.filename, equipment, args.names_matching)
  File "/home/user/workspace/TIP/gnpy/gnpy/core/network.py", line 38, in load_network
    return network_from_json(json_data, equipment)
  File "/home/user/workspace/TIP/gnpy/gnpy/core/network.py", line 63, in network_from_json
    el = cls(**el_config)
  File "/home/user/workspace/TIP/gnpy/gnpy/core/elements.py", line 389, in __init__
    if 'raman_pumps' in self.operational:
TypeError: argument of type 'NoneType' is not iterable
```
Can you please reproduce the clean Invocation error ?

AndreaDAmico:
> I have found another issue:
> before refactor if I try to launch an example file with a raman fiber without sim_param option:
> 
> ```
> (gnpy) user@esther-b:~/workspace/TIP/gnpy/examples$ python transmission_main_example.py ramantest.json 
> There are 76 channels propagating
> Power mode is set to True
> => it can be modified in eqpt_config.json - Span
> Invocation error: RamanFiber requires passing simulation params via --sim-params
> ```
> 
> after refactor:
> 
> ```
> (gnpy) user@esther-b:~/workspace/TIP/gnpy/examples$ python transmission_main_example.py ramantest.json 
> Traceback (most recent call last):
>   File "transmission_main_example.py", line 222, in <module>
>     network = load_network(args.filename, equipment, args.names_matching)
>   File "/home/user/workspace/TIP/gnpy/gnpy/core/network.py", line 38, in load_network
>     return network_from_json(json_data, equipment)
>   File "/home/user/workspace/TIP/gnpy/gnpy/core/network.py", line 63, in network_from_json
>     el = cls(**el_config)
>   File "/home/user/workspace/TIP/gnpy/gnpy/core/elements.py", line 389, in __init__
>     if 'raman_pumps' in self.operational:
> TypeError: argument of type 'NoneType' is not iterable
> ```
> 
> Can you please reproduce the clean Invocation error ?

I reproduce your error output feeding the transmission_main_example.py with a configuration .json  where the raman fiber json lacks of the operational field. I fix this introducing an if condition on this field in the RamanFiber constructor, even if I think an error should be raised if the operational field is not given in the constructor of RamanFiber. However, I do not think this implementation should be part of this restructuring PR, @EstherLerouzic and @jktjkt let me know what you think about it.

 The code already returns the Invocation error if no sim_params are given.

EstherLerouzic:
@jktjkt , The travis test are not passing but looking at the output of travis, it seems that this is related to the develop branch, because of test_science_util.py
```
==================================== ERRORS ====================================

_________________ ERROR collecting tests/test_science_utils.py _________________

tests/test_science_utils.py:14: in <module>

    from gnpy.core.network import load_sim_params

E   ImportError: cannot import name 'load_sim_params'

_________________ ERROR collecting tests/test_science_utils.py _________________

ImportError while importing test module '/home/travis/build/Telecominfraproject/oopt-gnpy/tests/test_science_utils.py'.

Hint: make sure your test modules/packages have valid Python names.

Traceback:

tests/test_science_utils.py:14: in <module>

    from gnpy.core.network import load_sim_params

E   ImportError: cannot import name 'load_sim_params'
```
But I have no error when running test on my computer ... Do you know where this error is coming from ?

jktjkt:
> @jktjkt , The travis test are not passing but looking at the output of travis, it seems that this is related to the develop branch, because of test_science_util.py
> [...]
> But I have no error when running test on my computer ... Do you know where this error is coming from ?

Looks like a result of a merge (tests for Raman were added in #320). This one will probably need rebasing; I think it's a nice opportunity for a cleanup as well.

jktjkt:
The discusssion continues in PR #329.

PR #327, opened 2019-12-10
Filtering feature

aleFerrari:
Add filter_penalty to Channel object to accumulate the filter penalties when the spectral information propagates.

### DIFF
--- a/gnpy/core/info.py
+++ b/gnpy/core/info.py
@@ -23,21 +23,14 @@ class Power(namedtuple('Power', 'signal nli ase')):
     ase: ase noise power in the signal -3 dB bandwidth (equal to the baud rate)"""
 
 
-
 class Channel(namedtuple('Channel', 'channel_number frequency baud_rate roll_off power filter_penalty')):
     """Object abstracting an optical signal:
-        channel_number: channel number
-        frequency: central frequency [Hz]
-        baud_rate: symbol rate [Baud]
-        roll_off: roll-off factor of root-raised-cosine shape, ∈ [0, 1]
-        power: Power object with signal, ase and nli power information
-        filter_penalty: an SNR penalty due to filtering [dB]"""
-
-    def __new__(cls, channel_number, frequency, baud_rate, roll_off, power, filter_penalty=None):
-        if not filter_penalty:
-            filter_penalty = 0
-
-        return super().__new__(cls, channel_number, frequency, baud_rate, roll_off, power, filter_penalty)
+    channel_number: channel number
+    frequency: central frequency [Hz]
+    baud_rate: symbol rate [Baud]
+    roll_off: roll-off factor of root-raised-cosine shape, ∈ [0, 1]
+    power: Power object with signal, ase and nli power information
+    filter_penalty: an SNR penalty due to filtering [dB]"""
 
 
 class Pref(namedtuple('Pref', 'p_span0, p_spani, neq_ch ')):
@@ -64,7 +57,7 @@ def create_input_spectral_information(f_min, f_max, roll_off, baud_rate, power, 
         pref=Pref(pref, pref, lin2db(nb_channel)),
         carriers=[
             Channel(f, (f_min+spacing*f),
-            baud_rate, roll_off, Power(power, 0, 0)) for f in range(1,nb_channel+1)
+            baud_rate, roll_off, Power(power, 0, 0), 0) for f in range(1,nb_channel+1)
             ])
     return si
 
--- a/gnpy/core/info.py
+++ b/gnpy/core/info.py
@@ -20,12 +20,17 @@ class Power(namedtuple('Power', 'signal nli ase')):
     """carriers power in W"""
 
 
-class Channel(namedtuple('Channel', 'channel_number frequency baud_rate roll_off power')):
-    pass
+class Channel(namedtuple('Channel', 'channel_number frequency baud_rate roll_off power filter_penalty')):
+
+    def __new__(cls, channel_number, frequency, baud_rate, roll_off, power, filter_penalty=None):
+        if not filter_penalty:
+            filter_penalty = 0
+
+        return super().__new__(cls, channel_number, frequency, baud_rate, roll_off, power, filter_penalty)
 
 
 class Pref(namedtuple('Pref', 'p_span0, p_spani, neq_ch ')):
-    """noiseless reference power in dBm: 
+    """noiseless reference power in dBm:
     p_span0: inital target carrier power
     p_spani: carrier power after element i
     neq_ch: equivalent channel count in dB"""
--- a/gnpy/core/info.py
+++ b/gnpy/core/info.py
@@ -17,10 +17,21 @@ from gnpy.core.utils import load_json
 from gnpy.core.equipment import automatic_nch, automatic_spacing
 
 class Power(namedtuple('Power', 'signal nli ase')):
-    """carriers power in W"""
+    """Power information of a channel in W:
+    signal: signal power
+    nli: non-linear interference power in the signal -3 dB bandwidth (equal to the baud rate)
+    ase: ase noise power in the signal -3 dB bandwidth (equal to the baud rate)"""
+
 
 
 class Channel(namedtuple('Channel', 'channel_number frequency baud_rate roll_off power filter_penalty')):
+    """Object abstracting an optical signal:
+        channel_number: channel number
+        frequency: central frequency [Hz]
+        baud_rate: symbol rate [Baud]
+        roll_off: roll-off factor of root-raised-cosine shape, ∈ [0, 1]
+        power: Power object with signal, ase and nli power information
+        filter_penalty: an SNR penalty due to filtering [dB]"""
 
     def __new__(cls, channel_number, frequency, baud_rate, roll_off, power, filter_penalty=None):
         if not filter_penalty:
@@ -37,6 +48,9 @@ class Pref(namedtuple('Pref', 'p_span0, p_spani, neq_ch ')):
 
 
 class SpectralInformation(namedtuple('SpectralInformation', 'pref carriers')):
+    """Object abstracting a WDM comb:
+    pref: noiseless reference power
+    carriers: list of Channel objects containing detailed information of each channel"""
 
     def __new__(cls, pref, carriers):
         return super().__new__(cls, pref, carriers)


ojnas:
Some comments:

- In general I think it's better to add things together with the feature using the thing, and not before.
- Introducing penalties is quite a big deviation from how gnpy has been functioning so far, where all impairments have been considered as noise and the end result is a (generalized) SNR. It will need some thought how to present the end result including penalties in a clear way.
- If we introduce penalties, it could be imagined that we want to model other impairments in a similar fashion in the future. Maybe it would be better to call the new attribute something more general like just `penalty` instead of `filter_penalty`?
- Setting a default value in the class instantiation seems a bit unnecessary. I think the Channel object is only ever used directly in the `create_input_spectral_information` function. Wouldn't it be cleaner to set a default value in that function? Maybe it could even be "hard-coded" to 0 there (as is the case for ASE and NLI power) since I find it difficult to imagine a case where the penalty would be already set to a non-zero value when the spectral information is first created.

aleFerrari:
> Some comments:
> 
> * In general I think it's better to add things together with the feature using the thing, and not before.

This pull request aims at giving a ground for the inclusion of filtering effects in GNPy and it follows a request of @jktjkt during the today call.
I understand your concerns but I think that it is better to first find an agreement on the structure of the 'thing' and then to start developing 'features' on top of it to avoid waste of time.

> * Introducing penalties is quite a big deviation from how gnpy has been functioning so far, where all impairments have been considered as noise and the end result is a (generalized) SNR. It will need some thought how to present the end result including penalties in a clear way.

Adding an SNR penalty is not opposite of using power levels (signal and noises) but complementary.
In fact, the use of only power levels is not enough to properly take into account filtering penalties.
Anyway I agree with you about the fact that it is fundamental to clearly present the overall results and I hope this will be a topic for discussion.

> * If we introduce penalties, it could be imagined that we want to model other impairments in a similar fashion in the future. Maybe it would be better to call the new attribute something more general like just `penalty` instead of `filter_penalty`?

I think it is better to have a clear name that explains the aim of the attribute and, in case we want to add new effects introducing an SNR penalty, they should have a new dedicated attribute.
In this case we can keep separate each physical effect, exactly as we have done for ASE and NLI: we do not have just one 'noise power' attribute but one for each type of noise.
I think this helps in having a clear structure and it allows a richer description of the propagation performances.

> * Setting a default value in the class instantiation seems a bit unnecessary. I think the Channel object is only ever used directly in the `create_input_spectral_information` function. Wouldn't it be cleaner to set a default value in that function? Maybe it could even be "hard-coded" to 0 there (as is the case for ASE and NLI power) since I find it difficult to imagine a case where the penalty would be already set to a non-zero value when the spectral information is first created.

Yes, I agree. Also @jktjkt gave similar suggestions. Thus, in the next days I'll try to modify the code in this direction.

ojnas:
> > * In general I think it's better to add things together with the feature using the thing, and not before.
> 
> This pull request aims at giving a ground for the inclusion of filtering effects in GNPy and it follows a request of @jktjkt during the today call.
> I understand your concerns but I think that it is better to first find an agreement on the structure of the 'thing' and then to start developing 'features' on top of it to avoid waste of time.
> 
It's fine, it was just a general comment. In these cases I think it would be better to mark the PR with something like "work in progress" and not merge it until the feature is ready. That way people can look at the proposed structure and discuss it without confusing others downloading the code with unused attributes. But it's up to @jktjkt to decide.
> > * If we introduce penalties, it could be imagined that we want to model other impairments in a similar fashion in the future. Maybe it would be better to call the new attribute something more general like just `penalty` instead of `filter_penalty`?
> 
> I think it is better to have a clear name that explains the aim of the attribute and, in case we want to add new effects introducing an SNR penalty, they should have a new dedicated attribute.
> In this case we can keep separate each physical effect, exactly as we have done for ASE and NLI: we do not have just one 'noise power' attribute but one for each type of noise.
> I think this helps in having a clear structure and it allows a richer description of the propagation performances.
> 
I see your point and think I agree. One option would be to have a penalty tuple (similar to the way we have a power tuple) including penalties from different effects. But maybe that can wait until we know if other effects will be modeled.

jktjkt:
> > Some comments:
> > 
> > * In general I think it's better to add things together with the feature using the thing, and not before.
> 
> This pull request aims at giving a ground for the inclusion of filtering effects in GNPy and it follows a request of @jktjkt during the today call.
> I understand your concerns but I think that it is better to first find an agreement on the structure of the 'thing' and then to start developing 'features' on top of it to avoid waste of time.

Thanks for sharing your changes early. I think that it's better to talk about changes in terms of actual diffs and patches here rather than slides during a VC :).

The way I understood your message during the call, I assumed that the propagation code was already written. If that's not the case, then I think it's better to not merge this PR as-is. Let's leave this one open, and I encourage you and your team to keep pushing your changes here as you work on the actual feature. That way you can get early feedback and we can only merge changes which are needed for the final result. I'll mark the PR with a work-in-progress label, or an *Request For Comment* prefix or something to make it obvious that this is, well, WIP.

## Workflow Recommendations

I would like this work to be done in a manner that is easy to review. Therefore I suggest to follow a workflow which involves force-pushes and history rewrites as we comment on the PR. For example, now there are three commits:
1) `Introduce filter_penalty attribute to Channel class with default value=0`
2) `Add description of SpectralInformation, Pref, Channel and Power`
3) `Remove default value of filter_penalty in Channel object`

The problem here is that 3) fixes the change done in 1), but given that there's been a documentation change in 2), you cannot just run a `git rebase --interactive origin/develop`, and move 3) to be before 2) right after 1) because of a merge conflict. So we need something better. I'll use `git checkout --patch` here to show how this history can be rewritten in response to the existing comments:

- switch to develop and make sure we're recent: `git checkout develop`, `git pull`
- start in a fresh branch: `git checkout -b wip-filtering`
- make sure that the content of this PR is avaialble in my git repo, `git fetch origin refs/pull/327/head`

Next, I'm going to interactively pick those changes which implement the real change, that addition of the `filter_penalty` variable. I'm going to ask git to offer me stuff from the **final** state of this branch as it is right now, and I'll be picking individual changes from there. The PR content is available via git's `FETCH_HEAD` symbolic reference, so I can run `git checkout --patch FETCH_HEAD`, and this is what I'm going to get:
```diff
diff --git b/gnpy/core/info.py a/gnpy/core/info.py
index 10a935f..1d2db67 100644
--- b/gnpy/core/info.py
+++ a/gnpy/core/info.py
@@ -17,21 +17,33 @@ from gnpy.core.utils import load_json
 from gnpy.core.equipment import automatic_nch, automatic_spacing
 
 class Power(namedtuple('Power', 'signal nli ase')):
-    """carriers power in W"""
+    """Power information of a channel in W:
+    signal: signal power
+    nli: non-linear interference power in the signal -3 dB bandwidth (equal to the baud rate)
+    ase: ase noise power in the signal -3 dB bandwidth (equal to the baud rate)"""
 
 
-class Channel(namedtuple('Channel', 'channel_number frequency baud_rate roll_off power')):
-    pass
+class Channel(namedtuple('Channel', 'channel_number frequency baud_rate roll_off power filter_penalty')):
+    """Object abstracting an optical signal:
+    channel_number: channel number
+    frequency: central frequency [Hz]
+    baud_rate: symbol rate [Baud]
+    roll_off: roll-off factor of root-raised-cosine shape, ∈ [0, 1]
+    power: Power object with signal, ase and nli power information
+    filter_penalty: an SNR penalty due to filtering [dB]"""
 
 
 class Pref(namedtuple('Pref', 'p_span0, p_spani, neq_ch ')):
-    """noiseless reference power in dBm: 
+    """noiseless reference power in dBm:
     p_span0: inital target carrier power
     p_spani: carrier power after element i
     neq_ch: equivalent channel count in dB"""
 
 
 class SpectralInformation(namedtuple('SpectralInformation', 'pref carriers')):
+    """Object abstracting a WDM comb:
+    pref: noiseless reference power
+    carriers: list of Channel objects containing detailed information of each channel"""
 
     def __new__(cls, pref, carriers):
         return super().__new__(cls, pref, carriers)
Apply this hunk to index and worktree [y,n,q,a,d,j,J,g,/,s,e,?]?
```
This screen works exactly as the `git add --patch` that you're already using. I decided that I fix the content first,. adding the docs later, so I start by splitting the hunks into smaller chunks (the '`s` key):

```diff
Split into 4 hunks.
@@ -17,6 +17,9 @@
 from gnpy.core.equipment import automatic_nch, automatic_spacing
 
 class Power(namedtuple('Power', 'signal nli ase')):
-    """carriers power in W"""
+    """Power information of a channel in W:
+    signal: signal power
+    nli: non-linear interference power in the signal -3 dB bandwidth (equal to the baud rate)
+    ase: ase noise power in the signal -3 dB bandwidth (equal to the baud rate)"""
 
 
Apply this hunk to index and worktree [y,n,q,a,d,j,J,g,/,e,?]?
```
nope, I don't need that one, it's just docs, therefore `n`.
```diff
@@ -21,7 +24,13 @@
 
 
-class Channel(namedtuple('Channel', 'channel_number frequency baud_rate roll_off power')):
-    pass
+class Channel(namedtuple('Channel', 'channel_number frequency baud_rate roll_off power filter_penalty')):
+    """Object abstracting an optical signal:
+    channel_number: channel number
+    frequency: central frequency [Hz]
+    baud_rate: symbol rate [Baud]
+    roll_off: roll-off factor of root-raised-cosine shape, ∈ [0, 1]
+    power: Power object with signal, ase and nli power information
+    filter_penalty: an SNR penalty due to filtering [dB]"""
 
 
 class Pref(namedtuple('Pref', 'p_span0, p_spani, neq_ch ')):
Apply this hunk to index and worktree [y,n,q,a,d,K,j,J,g,/,e,?]? 
```
I need just a part of this one, so I press `e` to bring me to my editor, and I make sure the content looks like this (essentially removing the doc string and making sure that the `pass` gets preserved):
```
 1 # Manual hunk edit mode -- see bottom for a quick guide.
 2 @@ -21,7 +24,13 @@
 3 
 4 
 5 -class Channel(namedtuple('Channel', 'channel_number frequency baud_rate roll_off power')):
 6 +class Channel(namedtuple('Channel', 'channel_number frequency baud_rate roll_off power filter_penalty')):
 7      pass
 8  
 9  
10  class Pref(namedtuple('Pref', 'p_span0, p_spani, neq_ch ')):

```
After a save+quit, I'm presented with the next hunk:
```diff
@@ -25,10 +34,10 @@
 
 
 class Pref(namedtuple('Pref', 'p_span0, p_spani, neq_ch ')):
-    """noiseless reference power in dBm: 
+    """noiseless reference power in dBm:
     p_span0: inital target carrier power
     p_spani: carrier power after element i
     neq_ch: equivalent channel count in dB"""
 
 
 class SpectralInformation(namedtuple('SpectralInformation', 'pref carriers')):
Apply this hunk to index and worktree [y,n,q,a,d,K,j,J,g,/,e,?]?
```
nope, that's a docs fix, so `n`.
```diff
@@ -29,9 +38,12 @@
     p_span0: inital target carrier power
     p_spani: carrier power after element i
     neq_ch: equivalent channel count in dB"""
 
 
 class SpectralInformation(namedtuple('SpectralInformation', 'pref carriers')):
+    """Object abstracting a WDM comb:
+    pref: noiseless reference power
+    carriers: list of Channel objects containing detailed information of each channel"""
 
     def __new__(cls, pref, carriers):
         return super().__new__(cls, pref, carriers)
Apply this hunk to index and worktree [y,n,q,a,d,K,j,J,g,/,e,?]?
```
`n`ope once again.
```diff
@@ -45,7 +57,7 @@ def create_input_spectral_information(f_min, f_max, roll_off, baud_rate, power,
         pref=Pref(pref, pref, lin2db(nb_channel)),
         carriers=[
             Channel(f, (f_min+spacing*f),
-            baud_rate, roll_off, Power(power, 0, 0)) for f in range(1,nb_channel+1)
+            baud_rate, roll_off, Power(power, 0, 0), 0) for f in range(1,nb_channel+1)
             ])
     return si
 
Apply this hunk to index and worktree [y,n,q,a,d,K,g,/,e,?]?
```
This one is valid, let's pick it by saying `y`.

After that, I got a bunch of unrelated bits, so I just `d` them.

Here's what has been staged so far, `git diff --staged`:
```diff
diff --git a/gnpy/core/info.py b/gnpy/core/info.py
index 10a935f..4489627 100644
--- a/gnpy/core/info.py
+++ b/gnpy/core/info.py
@@ -20,7 +20,7 @@ class Power(namedtuple('Power', 'signal nli ase')):
     """carriers power in W"""
 
 
-class Channel(namedtuple('Channel', 'channel_number frequency baud_rate roll_off power')):
+class Channel(namedtuple('Channel', 'channel_number frequency baud_rate roll_off power filter_penalty')):
     pass
 
 
@@ -45,7 +45,7 @@ def create_input_spectral_information(f_min, f_max, roll_off, baud_rate, power,
         pref=Pref(pref, pref, lin2db(nb_channel)),
         carriers=[
             Channel(f, (f_min+spacing*f),
-            baud_rate, roll_off, Power(power, 0, 0)) for f in range(1,nb_channel+1)
+            baud_rate, roll_off, Power(power, 0, 0), 0) for f in range(1,nb_channel+1)
             ])
     return si

```
That looks correct, so I run `pytest` to make sure there are no regressions, I make the commit and tell git to reuse metadata from the original commit in this series (`git log` to find out which one) via `git commit --reedit-message=790e9de4ecc3310cbea97c0d208d30086d2c6afa`. I fix the commit message in the editor to no longer mention any default value, and voila, here we go:
```
[wip-filtering a1098d0] Introduce filter_penalty attribute to Channel class
 Author: Alessio Ferrari <alessio.ferrari@polito.it>
 Date: Tue Dec 10 12:27:12 2019 +0100
 1 file changed, 2 insertions(+), 2 deletions(-)
```
Let's pick the remaining changes via running `git checkout --patch FETCH_HEAD` again:
```diff
diff --git b/gnpy/core/info.py a/gnpy/core/info.py
index 4489627..1d2db67 100644
--- b/gnpy/core/info.py
+++ a/gnpy/core/info.py
@@ -17,21 +17,33 @@ from gnpy.core.utils import load_json
 from gnpy.core.equipment import automatic_nch, automatic_spacing
 
 class Power(namedtuple('Power', 'signal nli ase')):
-    """carriers power in W"""
+    """Power information of a channel in W:
+    signal: signal power
+    nli: non-linear interference power in the signal -3 dB bandwidth (equal to the baud rate)
+    ase: ase noise power in the signal -3 dB bandwidth (equal to the baud rate)"""
 
 
 class Channel(namedtuple('Channel', 'channel_number frequency baud_rate roll_off power filter_penalty')):
-    pass
+    """Object abstracting an optical signal:
+    channel_number: channel number
+    frequency: central frequency [Hz]
+    baud_rate: symbol rate [Baud]
+    roll_off: roll-off factor of root-raised-cosine shape, ∈ [0, 1]
+    power: Power object with signal, ase and nli power information
+    filter_penalty: an SNR penalty due to filtering [dB]"""
 
 
 class Pref(namedtuple('Pref', 'p_span0, p_spani, neq_ch ')):
-    """noiseless reference power in dBm: 
+    """noiseless reference power in dBm:
     p_span0: inital target carrier power
     p_spani: carrier power after element i
     neq_ch: equivalent channel count in dB"""
 
 
 class SpectralInformation(namedtuple('SpectralInformation', 'pref carriers')):
+    """Object abstracting a WDM comb:
+    pref: noiseless reference power
+    carriers: list of Channel objects containing detailed information of each channel"""
 
     def __new__(cls, pref, carriers):
         return super().__new__(cls, pref, carriers)
Apply this hunk to index and worktree [y,n,q,a,d,s,e,?]?
```
That's all right, so I say `y`. Then I'm prompted for the rest (that's an artifact of the fact that I started on current `develop` which is d483802 instead of starting my `wip-filtering` branch on 9a0eece69c8491de069458a1575ae40035830a12, oops, I made a mistake there. Let's don't bother with that, I can always `q` from my editor to skip any other prompting and discard the remaining hunks. Again, let's test via `pytest` and commit, now referencing the second original commit:
```
jkt@krno ~/work/TIP/oopt-gnpy[wip-filtering*] $ git commit --reedit-message=38380b988e3d60c799fdaa8a545fa2151d8854d5
[wip-filtering 68d193e] Add description of SpectralInformation, Pref, Channel and Power
 Author: Alessio Ferrari <alessio.ferrari@polito.it>
 Date: Tue Dec 10 12:45:43 2019 +0100
 1 file changed, 15 insertions(+), 3 deletions(-)
```
### Now I'm done...
...and the history is nice and reviewable. I can force-push this current branch to the original location where the PR was created from, so in this case I do `git push --force name_of_my_for HEAD:filtering_feature`.

jktjkt:

PR #328, opened 2019-12-12
User error catching improvment

EstherLerouzic:
Give more useful comment for user to correct topology
    
    when a node name used in link does not correspond to listed
    node names, Bad link msg was thrown without info on which link
    causes the problem.
    This change gives additional info on link and catch the user error
    more cleanly.


gnpy/core/convert.py
```diff
@@ -511,7 +512,10 @@ def parse_excel(input_filename):
         raise ValueError(f'Duplicate city: {all_cities}')
     if any(ln.from_city not in all_cities or
            ln.to_city   not in all_cities for ln in links):
-        raise ValueError(f'Bad link.')
+        for lnk in links:
```
jktjkt: This checks these links twice, and the error message is printed to stdout. I think it would be better to put it into the exception instead.

I think it would be OK to just get rid of the initial `any(...)` check, perform the explicit loop as this patch is doing, and upon seeing a first failure:
```py
raise NetworkTopologyError(f"Bad link from {lnk.from_city} to {ln.to_city}')
```

EstherLerouzic: But then it stops at the first error, and the user has to rerun the program. I would prefer to have the list of all wrong names when I debug my topologies...

EstherLerouzic:
Yes I made a mistake, and this pR was already open in september :)

gnpy/core/convert.py
jktjkt: I still think that encoding an error in an exception is preferred over directly printing it. That way, the info is available to other users than just those using console. Can you just sweep through the list and file a temporary as needed, and then put that info into the exception?

That said, this is "just" an XLS convertor and the REST API won't work with XLS files (I hope!), so my negative vote is a suggestion, not a request.

jktjkt:
(Reopening, I think that this got closed by accident. It's an old commit, sure, but I think it's a good change, so let's make sure it won't get lost. Esther, feel free to close if you really meant to close.)

EstherLerouzic:
Since I opened two PR for the same code (sorry for the mistake) can we continue discussion on #304

jktjkt:
Ah, my bad. I missed #304.

PR #329, opened 2019-12-17
Polito's Fiber paramter refactoring

jktjkt:
This is a cleaned-up version of #319; the cleanup/squashing was done by @EstherLerouzic. I went over the result and tried to give proper credits to all contributors, hence the `Co-authored-by` headers.

### DIFF
--- a/gnpy/core/science_utils.py
+++ b/gnpy/core/science_utils.py
@@ -1,6 +1,5 @@
 import numpy as np
 from operator import attrgetter
-from collections import namedtuple
 from logging import getLogger
 import scipy.constants as ph
 from scipy.integrate import solve_bvp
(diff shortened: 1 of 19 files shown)


gnpy/core/science_utils.py
```diff
@@ -270,42 +132,68 @@ def raised_cosine_comb(f, *carriers):
...
+        cls._shared_dict['sim_params'] = sim_params
+
+    @classmethod
+    def get_simulation(cls):
+        self = cls.__new__(cls)
+        return self
+
+    @property
+    def sim_params(self):
+        return self._shared_dict['sim_params']
```
jktjkt: The code now "hides" the shared state via a class' static variable. That's functionally equivalent to a global variable, albeit with a better namespace. As far as I can tell, this is not explained in the commit messages of the original PR. I wonder why did you elect for this approach as opposed to explicitly passing the simulation parameters to all consumers?

jktjkt: Cc @aleFerrari , @andreadamico -- not sure if you're getting mails from this PR where you are not the PR "creator" in GitHub.

jktjkt: During today's call we agreed @AndreaDAmico will provide a pseudocode example on how Polito plans to use this feature. After that, we will check whether it's feasible to implement this via explicit state passing.

AndreaDAmico: We would like to achieve an implementation that enables multiple evaluations, each with different simulation parameters, performed on the same network instance. In the current version of GNPy, the simulation parameters are stuck to the fiber instances as the network is constructed, preventing the above mentioned usage. In the new implementation we have defined a mutable 'singleton' that can be retrieved by the final consumer in an implicit way.    
In any case, a pseudocode example is the following:

import gnpy
network_instance = gnpy.build_network(topology_description)

set_simulation_parameters( sim_params_1)

infos_1 = propagate( path, req, equipment)

set_simulation_parameters( sim_params_2)

infos_2 = propagate( path, req, equipment)

Issue #334, opened 2020-01-16
power saturation is not correctly handled in autodesign when mixt with user defined values


EstherLerouzic:
The power limitation (saturation) is only applied during propagation so that user imposed values of gain are considered instead of saturated gain due to power limitation in the delta power computation used for optimisation on following elements.
This issue should be handled in a further correction.

PR #335, opened 2020-01-16
Add gain limitation

EstherLerouzic:
Creating this Pull request for you to see my current work (and have reports).

When mixing autodesign with user defined amplifiers value, some inconsistency on gain were not addressed, so that the gain on the user specified amplifiers was not checked to be below acceptable max gain. This work corrects this:
- if the specified gain is above max_gain + extended gain range, a warning is raised, and the gain is limited to max_gain + extended gain range.
- else there is no change
The correction brought with this change enabled me to discover another issue #334


EstherLerouzic:
This one has to be corrected

PR #337, opened 2020-02-13
Correction bug on include node constraint

EstherLerouzic:
The check on node_list constraint has to be be done on all elements except the last one,
and the last one is always a "STRICT" value. this means that if no constraints are given
this list is equals to ["STRICT"]. Previous code was using len("STRICT") ie 6 instead of
len(["STRICT"]) ie 1.

### DIFF
--- a/gnpy/core/request.py
+++ b/gnpy/core/request.py
@@ -291,7 +291,7 @@ def compute_constrained_path(network, req):
               'be destination trx'
         LOGGER.critical(msg)
         raise ValueError()
-
+    # len(nodes_list) is 1 or more since it must include destination node
     if len(nodes_list) == 1:
         try:
             total_path = dijkstra_path(network, source, destination, weight='weight')
@@ -311,6 +311,7 @@ def compute_constrained_path(network, req):
             req.blocking_reason = 'NO_PATH'
             total_path = []
     else:
+        # len(nodes_list) is 2 or more (includes at list one include node apart from destination)
         all_simp_pths = list(all_simple_paths(network, source=source,\
             target=destination, cutoff=120))
         candidate = []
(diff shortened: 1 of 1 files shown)


gnpy/core/request.py
```diff
@@ -330,13 +330,13 @@ def compute_constrained_path(network, req):
             # last node which is the transceiver)
             # if all nodes i n node_list are LOOSE constraint, skip the constraints and find
             # a path w/o constraints, else there is no possible path
-            if nodes_list[:-len("STRICT")]:
+            if nodes_list[:-len(['STRICT'])]:
```
ojnas: I'm a bit confused by this code. Firstly, since `len(['STRICT'])` will always be 1, why not just write 1? Secondly, this just checks if there are more than one element in the list, right? But you have already checked if length of `nodes_list` is exactly 1 above and this code is in the else clause so it will always be true. Am I missing something?

EstherLerouzic: @ojnas thanks for the feedback ! Indeed the bug comes from an initial error I made using -6 instead of -1, and len('STRICT') instead of 6. nodes_list is the ordered list of include nodes specified by the user + the destination node.
So this first test checks if there are include nodes in the list.  At this stage of the program, there is no path found, and the purpose of this test is to check if there was some constraint added by the user.

- if nodes_list[:-1] not empty -> test is true: the message informs the user about the impossibility to find a path with the list of include nodes
- If nodes_list[:-1] is empty the test is False, and the message should inform the user that there is no path found: so you are right the second message is misleading.

then the second set of tests verifies if the include node list is strict or not (again removing the last element). If no "STRICT" is found in the list then  the constraints are not an obligation, and another path is searched without them.

Is this clearer ?
I can try to improve the test and messages  or do you have any better suggestion ?

EstherLerouzic: I have corrected according to your suggestions. is it better ?

jktjkt:
I've squashed these two commits together and I'll be uploading them into another PR (together with modified commits from #339) because I do not have permission to overwrite this PR.

-> uploaded as #342.

PR #339, opened 2020-02-19
Add test include constraint

EstherLerouzic:
This PR adds tetst according to #337


jktjkt:
history rewritten in #342

PR #377, opened 2021-01-02
Rollback numpy to 1.19.3 for temporary Window Runtime Error

mohammadsaadraza:
# Thanks for contributing to GNPy

If it isn't much trouble, please send your contribution as patches to our Gerrit.
Here's [how to submit patches](https://review.gerrithub.io/Documentation/intro-gerrit-walkthrough-github.html), and here's a [list of stuff we are currently working on](https://review.gerrithub.io/p/Telecominfraproject/oopt-gnpy/+/dashboard/main:main).
Just sign in via your existing GitHub account.

However, if you feel more comfortable with filing GitHub PRs, we can work with that too.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ alabaster>=0.7.12,<1
 docutils==0.15.2
 matplotlib>=3.1.0,<4
 networkx>=2.3,<3
-numpy>=1.19.4,<2
+numpy>=1.19.3,<2
 pandas>=1.0.5,<2
 pbr>=5.4.4,<6
 Pygments>=2.4.2,<3


jktjkt (changes requested):
Hi @mohammadsaadraza , can you please clarify what issue this patch fixes? The commit message mentions "temporary Window Runtime Error", but I don't know what that means. Please provide more info in the commit message, and let us know.

In case you have not tried our patch submission via Gerrit, I encourage you to please [submit your patch in there](https://review.gerrithub.io/Documentation/intro-gerrit-walkthrough-github.html) as it makes the workflow a bit more streamlined. Thanks!

mohammadsaadraza:
@jktjkt numpy==1.19.4, dependency of GNPy, currently give a Runtime Error of Windows Platforms. Changing it to numpy==1.19.3 is a temporary solution until Windows Updates resolve that error. 

So I just changed numpy version 1.1.9.4 to 1.19.3 in "requirements.txt" file and installed GNPy again. It is a temporary fix and it allows me to try GNPy on my Windows.

ojnas:
See also:
https://github.com/numpy/numpy/issues/17726
and:
https://developercommunity.visualstudio.com/content/problem/1207405/fmod-after-an-update-to-windows-2004-is-causing-a.html

Since the original issue is a bug in Windows and the fix in numpy 1.19.3 seems to cause other problems:
https://github.com/numpy/numpy/releases/tag/v1.19.4,
I don't think we should do this rollback.

mohammadsaadraza:
I wasn't able to run GNPy at all with numpy==1.19.4

I just made the review request to make contributors to GNPy aware that numpy==1.19.3 does temporarily solve this problem for me on Windows. I was able to run some of the basic features and tests. 

The GNPy dependency of numpy==1.19.4 only is, I believe, a little strict. Just an observation though. The contributors can do as they see fit.

jktjkt:
Hi @mohammadsaadraza , thanks for clarifying this. Note that we actually depend on `numpy>=1.19.4`, we are not pinning an exact version via `==`.

This is apparently caused by a bug in one of the latest updates to Windows. Considering that Microsoft apparently plans to release a fix by the end of Januray 2021 (suggested by [Steve Wishnousky](https://developercommunity.visualstudio.com/content/problem/1207405/fmod-after-an-update-to-windows-2004-is-causing-a.html) in a Microsoft website which doesn't support linking to individual posts) and that the workaround is `pip install numpy==1.19.3` which is reported as broken on some Linux systems, I think there's no perfect solution from our side; we are just consumers of what the platform gets us I'm afraid.

Thanks for reporting this!

PR #410, opened 2021-06-15
CI: test on Windows

jktjkt:


gnpy/tools/cli_examples.py
```diff
@@ -306,7 +306,11 @@ def path_requests_run(args=None):
     _setup_logging(args)
 
     _logger.info(f'Computing path requests {args.service_filename} into JSON format')
-    print(f'{ansi_escapes.blue}Computing path requests {os.path.relpath(args.service_filename)} into JSON format{ansi_escapes.reset}')
```
ojnas, line 309: We could also just remove this relatively pointless printout.

jktjkt: Right, that's much easier :), thanks.

PR #455, opened 2022-12-06
Customized branch for use in LYNX (leave PR open, need not merge)

ghost:
Some changes are as follows:

- a different format for response generation
- comment out console prints
- another json function for elements, to view metrics as the signal is propagated.

This is an opinionated customization specifically for LYNX and will probably include more changes as we need them. Any suggestions are welcome.

### DIFF
--- a/gnpy/core/network.py
+++ b/gnpy/core/network.py
@@ -151,7 +151,7 @@ def target_power(network, node, equipment):  # get_fiber_dp
 
     SPAN_LOSS_REF = 20
     POWER_SLOPE = 0.3
-    dp_range = list(equipment['Span']['default'].delta_power_range_db)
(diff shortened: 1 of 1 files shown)


jktjkt:
Hi @mohammadsaadraza , awesome to see how you're using GNPy. When you're ready to push small, incremental changes to GNPy, please consider using Gerrit since that's where the development is taking place.

Since this looks like a long-term work-in-progress draft PR, I wonder if there's a technical reason for leaving it open. Does it make your workflow easier in some way? If not, I would prefer to close the PR since it doesn't appear that we (GNPy maintainers) can do much at this point.

mohammadsaadraza:
Thank you Jan. The reason I requested that PR be left open is for the
convenience of someone else looking to use GNPy in a server/API environment.

Ofcourse it is still an opiniated customized version meant for LYNX, but i
guess it can help someone else as well.

And that this GNPy repo has more traffic than ours. Maybe in the future,
GNPy maintainer could officially release a separate version using this PR.

PR #560, opened 2026-02-07
Add total_loss parameter for fiber attenuation (OTDR measurement support)

### LINKED ISSUE #441: Fiber attenuation: total loss vs. loss-per-km
Today, fiber loss is specified in the IO format indirectly, via the `length` and the `loss_coef`. Vinay raised a good point in that users typically have access to OTDR measurements, and these report a total loss value. Should we switch to working with the total loss of the fiber (except the connectors) in the YANG interface? That way, people won't have to divide that number from the OTDR measurement by the fiber length.

The format currently supports specifying the loss either as a scalar value (which applies over the whole spectrum), or as a mapping of `frequency`→ `loss_coeff`; this will have to be preserved either way.

Cc: @sme791

LynchXLQ:
## Summary

This PR adds support for specifying fiber attenuation as **total loss (dB)** instead of only via `loss_coef` (dB/km). This addresses the use case where users have OTDR measurements that report total span loss directly.

Resolves #441

### `gnpy/core/parameters.py` (FiberParams)
- Accept `total_loss` as an alternative to `loss_coef`
- When `total_loss` is provided, derive `loss_coef = total_loss / length`
- Support both scalar and frequency-dependent `total_loss` (dict with `value` and `frequency` keys)
- Fully backward compatible: `loss_coef` continues to work unchanged
- `asdict()` preserves the original parameter (`total_loss` or `loss_coef`)

### `gnpy/core/elements.py` (Fiber)
- `to_json` exports `total_loss` / `total_loss_per_frequency` when total_loss was the input

### `gnpy/tools/yang_convert_utils.py` & `convert_legacy_yang.py`
- YANG format conversion support for `total_loss` / `total_loss_per_frequency`

### Tests
- Scalar total_loss → correct loss_coef derivation
- Frequency-dependent total_loss → correct per-frequency loss_coef
- Backward compatibility (loss_coef still works)
- Round-trip serialization via `asdict()`

## Usage Example

```json
{
  "uid": "fiber-span-1",
  "type": "Fiber",
  "params": {
    "length": 80,
    "length_units": "km",
    "total_loss": 16.0,
    "con_in": 0.5,
    "con_out": 0.5,
    "pmd_coef": 1.265e-15
  }
}
```

cc @jktjkt

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -18,6 +18,7 @@ To learn how to contribute, please see CONTRIBUTING.md
 - Gabriele Galimberti (Cisco) <ggalimbe@cisco.com>
 - Gert Grammel (Juniper Networks) <ggrammel@juniper.net>
 - Giacomo Borraccini (NEC Laboratories America) <gborraccini@nec-labs.com>
+- Linqi Xiao (University of Texas at Dallas) <linqi.xiao@utdallas.edu>, <linqixiao2021@gmail.com>, <linqi.xiao@ieee.org>
 - Gilad Goldfarb (Facebook) <giladg@fb.com>
 - James Powell (Telecom Infra Project) <james.powell@telecominfraproject.com>
 - Jan Kundrát (Telecom Infra Project) <jkt@jankundrat.com>
--- a/gnpy/core/elements.py
+++ b/gnpy/core/elements.py
@@ -903,7 +903,14 @@ class Fiber(_Node):
             'con_in': self.params.con_in,
             'con_out': self.params.con_out
         }
-        if isinstance(self.params.loss_coef, ndarray):
+        if self.params.total_loss is not None:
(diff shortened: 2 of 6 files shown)


EstherLerouzic:
Dear @LynchXLQ , thanks a lot for your contribution! We will review it and will give you feedback as soon as possible. We normally use gerrithub for review (https://review.gerrithub.io/q/project:Telecominfraproject/oopt-gnpy). If you are familiar with it please push your commits there. If not I can do it for you, so that we can start the review process.
Regards

LynchXLQ:
Thanks @EstherLerouzic! I've pushed the changes to GerritHub: https://review.gerrithub.io/c/Telecominfraproject/oopt-gnpy/+/1231186

Looking forward to your feedback!

